PR #5, opened 2023-08-27
[formatter] fish_indent

MahanRahmati:
This PR adds fish_indent from [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/fish_indent.lua).


stevearc:
Thanks for the PR!

PR #14, opened 2023-08-30
feat: add shellharden

WhoIsSethDaniel:


stevearc:
Thanks!

WhoIsSethDaniel:
This isn't actually working, but I need some advice.  The correct command to have `shellharden` do its job is:
```
cat <file> | shellharden --transform ''
```
Note the two quotes at the end. The result is a corrected shell file sent to stdout.

However, I can't get the two quotes to execute properly. The log has this:
```
[INFO] Run shellharden on /home/seth/.bash-init/vi.sh
[DEBUG] Run command: { "shellharden", "--transform", "''" }
[ERROR] shellharden exited with code 1
[WARN] shellharden stdout: { "" }
[WARN] shellharden stderr: { "'': No such file or directory (os error 2)", "" }
[ERROR] Formatter 'shellharden' error: '': No such file or directory (os error 2)
```

WhoIsSethDaniel:
Ooh. Sorry. I should have submitted this as a draft. Please see the  previous comment.

stevearc:
I believe what you would want then is
```lua
args = { "--transform", "" },
```
Can you give that a try and see if it works?

PR #24, opened 2023-09-03
fix readme error and add a tip

jinzhongjia:
fix config error, and add a tip for formatters list

Oh, it is `prettierd`, not `prettier_d`, it gives is that "No config found". At first I thought it was because the configuration of pretter was not given.

Then I started to troubleshoot my configuration file, and finally I found out that the sample configuration in the readme was wrong 🥹

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ require("conform").setup({
     formatters_by_ft = {
         lua = { "stylua" },
         -- Conform will use the first available formatter in the list
-        javascript = { "prettierd", "prettier" },
+        javascript = { "prettier_d", "prettier" },
         -- Formatters can also be specified with additional options
         python = {
             formatters = { "isort", "black" },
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -147,6 +147,17 @@ See [conform.format()](#formatopts-callback) for more details about the paramete
...
+
+```lua
+-- for prettier_d
+require("conform").setup({
+    formatters_by_ft = {
+        javascript = { "prettierd"},
+    },
+})
+```
+
```
stevearc, line 160: The list of formatters is literally right below this, and the code example seems duplicative of the other examples. I don't think this note is necessary.

stevearc (changes requested):
Thanks for the fix! The documentation in the README and help file are actually generated, so you'll need to make the change here:
https://github.com/stevearc/conform.nvim/blob/ff7f8fdc0946d546a5a3f0617dbbf10f660dbb58/scripts/options_doc.lua#L6

README.md
jinzhongjia: ok, I will remove it

stevearc:
Thanks!

PR #28, opened 2023-09-06
fix: format() takes bufnr, not buf

WhoIsSethDaniel:


WhoIsSethDaniel:
It looks like there was a buf -> bufnr format() arg I missed earlier.

stevearc:
Thanks!

PR #29, opened 2023-09-07
feat: add beautysh, taplo, trim_newlines and trim_whitespace

xfzv:


lua/conform/formatters/trim_newlines.lua
```diff
@@ -0,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/trim_newlines.lua",
+    description = "Trim new lines",
```
stevearc, line 5: For this and the one below, mention that it's using `awk` in the description and change the link to a landing page for awk. Maybe https://www.gnu.org/software/gawk/manual/gawk.html ?

stevearc (changes requested):
Thanks for the additions! Small request for the description/url of the awk commands

PR #43, opened 2023-09-14
feat: add buf as protobuf linter

buztard:


stevearc (changes requested):
Does buf support stdin? Just looking at the [cli docs](https://buf.build/docs/reference/cli/buf/format#debug) it doesn't mention it anywhere. We may need to use the `-w` flag and set `stdin = false` in the definition.

buztard:
Oh, you're right, it ignores stdin and scans the whole package instead. My bad

stevearc:
Thanks for the PR!

PR #48, opened 2023-09-15
fix(health): use non-deprecated functions if available

MariaSolOs:
The `report_` health functions have been deprecated, and so this PR updates the health check to use the newer versions of the reporters if available.


stevearc:
Thanks for the PR!

PR #65, opened 2023-09-20
Change buf format URL

stefanvanburen:
Noticed this was pointing to the `buf lint` docs - seemed like the `buf format` CLI reference might be better. An alternative might be the `buf format` [style reference](https://buf.build/docs/format/style), but that doesn't seem as good.


stevearc:
Yeah that link seems more relevant. Thanks!

PR #66, opened 2023-09-21
Add ruff formatter and improve ruff root finding

leiserfg:


stevearc:
Thanks for the PR!

lua/conform/formatters/ruff.lua
```diff
@@ -15,5 +15,6 @@ return {
   stdin = true,
   cwd = require("conform.util").root_file({
     "pyproject.toml",
+    "ruff.conf",
```
leiserfg, line 18: Sorry there was a typo here, it should be `ruff.toml`, same in the other formatted.

PR #68, opened 2023-09-21
feat: add mdformat

rnprest:


rnprest:
I couldn't find a `CONTRIBUTING.md`, so I just looked at what typical changes others have done for adding new formatters - in this case, just adding an `mdformat.lua` and adding the 2 doc lines. Please let me know if I missed anything!

Also, not quite sure how to get range formatting to work for this?

stevearc:
Thanks for the PR! It doesn't look like mdformat accepts any arguments for range formatting, so in this case you don't need to do anything extra; conform will automatically do its own "best effort" at range formatting.

PR #70, opened 2023-09-22
fix: `stylelint` and `markdownlint` when there are non-autofixable errors

chrisgrieser:


stevearc:
Thanks for the fixes!

PR #75, opened 2023-09-25
feat: add `markdown-toc`

chrisgrieser:


lua/conform/formatters/markdown-toc.lua
```diff
@@ -0,0 +1,15 @@
...
+  meta = {
+    url = "https://github.com/jonschlinkert/markdown-toc",
+    description = "API and CLI for generating a markdown TOC (table of contents) for a README or any markdown files.",
+  },
+  command = "markdown-toc",
+  stdin = false,
+  args = function()
+    -- use the indentation set in the current buffer, effectively allowing us to
+    -- use values from .editorconfig
+    local indent = vim.bo.expandtab and (" "):rep(vim.bo.tabstop) or "\t"
```
stevearc, line 12: nit: technically formatting can happen on the non-current buffer (for example, if you're in another file and `:wa`). To handle this, use `funtion(ctx)` and `vim.bo[ctx.buf]`

stevearc (changes requested):
One nit

stevearc:
Thanks for the PR!

PR #78, opened 2023-09-26
feat(php-cs-fixer): Add support for php-cs-fixer

aidenlangley:
- Added a simple utility to check a list of paths and return the first that exists, or a default. Useful for defining commands for a formatter so it will call the correct executable.
- Added support for `php-cs-fixer`.


stevearc (changes requested):

aidenlangley:
Using `find_executable` now instead. Hope the history isn't too messy.

stevearc:
Thanks for the PR!

PR #79, opened 2023-09-26
fix(golines): Pass $FILENAME to golines

aidenlangley:
`golines` with no arguments tries to format the entire project which takes too long and times out.

### DIFF
--- a/lua/conform/formatters/golines.lua
+++ b/lua/conform/formatters/golines.lua
@@ -5,4 +5,5 @@ return {
     description = "A golang formatter that fixes long lines",
   },
   command = "golines",
+  args = { "$FILENAME" },
 }


stevearc:
According to [the usage docs](https://github.com/segmentio/golines#usage)

> The paths can be either directories or individual files. If no paths are provided, then input is taken from `stdin` (as with `gofmt`).

Are you sure that's what's actually happening? I ran it with strace and don't see it opening any other files, much less the whole project.

aidenlangley:
It just hangs and does nothing, expecting stdin input.

aidenlangley:
No, my mistake, it actually hangs and does nothing, expecting something from stdin.

aidenlangley:
Sorry about the duplicate comment there, using my phone.

This is the project I was working and testing on:
https://git.sr.ht/~nedia/nedots.go

Not sure exactly what version but likely latest of both, I'll verify when I can.

aidenlangley:
```sh
❯ nvim -v
NVIM v0.9.2
❯ golines --version
golines v0.11.0
```

I can't seem to reproduce the issue today. Admittedly I wrote the fix and then inferred what was going on. I've run into this on a couple of occasions where `conform` will report that `golines` has timed out, so the next time it crops up I'll investigate further with log level to trace & `:ConformInfo` and open an issue with some real info.

PR #80, opened 2023-09-27
feat: add darker

okuuva:
[`darker`](https://github.com/akaihola/darker) runs [`black`](https://github.com/psf/black) only on the changed lines in a git repo. Useful in situations where the whole repo hasn't been painted black for one reason or another but you still want to be able to run autoformatter on changes you made.

### DIFF
--- a/lua/conform/formatters/darker.lua
+++ b/lua/conform/formatters/darker.lua
@@ -6,7 +6,15 @@ return {
     description = "Run black only on changed lines.",
   },
   command = "darker",
-  args = { "--stdout", "--quiet", "$FILENAME" },
+  args = {
+    "--quiet",
+    "--no-color",
+    "--stdout",
+    "--revision",
+    "HEAD..:STDIN:",
+    "--stdin-filename",
+    "$FILENAME",
+  },
   cwd = util.root_file({
     -- https://github.com/akaihola/darker#customizing-darker-black-isort-flynt-and-linter-behavior
     "pyproject.toml",
--- a/lua/conform/formatters/darker.lua
+++ b/lua/conform/formatters/darker.lua
@@ -1,0 +1,14 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/akaihola/darker",
+    description = "Run black only on changed lines.",
+  },
+  command = "darker",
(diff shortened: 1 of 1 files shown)


lua/conform/formatters/darker.lua
```diff
@@ -0,0 +1,14 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/akaihola/darker",
+    description = "Run black only on changed lines.",
+  },
+  command = "darker",
+  args = { "--stdout", "--quiet", "$FILENAME" },
```
stevearc, line 9: I think you're going to want to use `--stdin-filename`

stevearc (changes requested):

lua/conform/formatters/darker.lua
okuuva: Huh. That's true. The weird thing is that I've been using `darker` first with `null-ls` and now with `conform.nvim` as a custom engine/formatter for quite a while now with those args and it's been working fine. Maybe by accident 😅 . Will fix!

okuuva: Hmm, I'm actually not sure if we want to use `--stdin-filename` here. Without it `darker` respects the revision option set in pyproject.toml if it's there so this way files formatted with `darker` via `conform` would match the files formatted with `darker .` in project root. Even if the root isn't a git repo.

stevearc: The trouble is what happens if you run this formatter on a buffer that has unsaved changes? It will read the current state from disk, which will blow away your changes. I assume you're using this formatter in a `BufWritePost` hook?

What do you think about doing something like this:
```lua
args = function(ctx)
  if vim.bo[ctx.buf].modified then
    return { "--stdout", "--quiet", "--stdin-filename", "$FILENAME" }
  else
    return { "--stdout", "--quiet", "$FILENAME" }
  end
end
```

okuuva: Oh yeah, I hadn't realized that auto-save.nvim had saved me from losing changes even when I had the formatter configured as `format_on_save`. Yeah, context aware args sound like a best-of-both-worlds solution to me. I'll update the PR shortly. Thank you for your patience and teaching me something new ^^

stevearc:
Thanks for the PR!

PR #83, opened 2023-09-28
feat: format injected languages

stevearc:
New `injected` formatter that will find injected languages using treesitter and format those code snippets with the configured formatters. To use:

```lua
require("conform").setup({
  formatters_by_ft = {
    markdown = { "injected" },
  }
})
```


chrisgrieser:
was actually thinking to PR support for `cbformat`, but his is actually way nicer. Thanks for this one!

PR #100, opened 2023-10-02
fix(injected): parse injections

folke:
On nightly, the injected formatter doesn't work.

`parse(true)` needs to be used to parse injections.
See also https://github.com/folke/noice.nvim/pull/571

Caused by upstream change: https://github.com/neovim/neovim/commit/2ca076e45fb3f1c08f6a1a374834df0701b8d778

### DIFF
--- a/lua/conform/formatters/injected.lua
+++ b/lua/conform/formatters/injected.lua
@@ -62,7 +62,11 @@ return {
       callback("No treesitter parser for buffer")
       return
     end
-    parser:parse()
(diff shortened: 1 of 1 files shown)


folke:
Disabled the diagnostic to pass the typecheck github action.
This new param is available on nightly, but not on stable
 Stable doesn't have any parameters, so it's safe to always pass `true`

stevearc:
Agh, of course nightly. Thanks for the fix!

folke:
Thanks for merging! :)

PR #106, opened 2023-10-03
feat: add ts-standard

pyrho:
Thanks for the plugin ! 

Cheers.

edit: actually it's not working yet 😬 , converting to draft until I figure it out.


pyrho:
There's currently an issue with standard which prevents feeding files from stdin, see https://github.com/standard/ts-standard/issues/273 .
Discarding this for now. Sorry for the noise.

stevearc:
@pyrho did you try `stdin = false`? Conform supports formatters that don't accept stdin by writing a temp file and formatting it in-place.

pyrho:
Hey, just tried that but then there's another issue; it seems like `ts-standard` ignore hidden files by default.

```
File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override
```

I've tried all the combinations I could think of for the ignore pattern but I can't make it work, this message actually comes from `eslint` (which is used by ts-standard) so my gut tells me that `ts-standard` does not propagate the command line option all the way to eslint.

ts-standard feels a bit rough around the edges right now, might be worth revisiting in a year or two :)

In any case, thank you for conform!

PR #107, opened 2023-10-04
add support for mix format

ethangunderson:
I think this one is pretty straight forward to add. I've tested this locally but let me know if I missed anything.

### DIFF
--- a/lua/conform/formatters/mix.lua
+++ b/lua/conform/formatters/mix.lua
@@ -1,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://hexdocs.pm/mix/main/Mix.Tasks.Format.html",
+    description = "Format Elixir files using the mix format command.",
+  },
+  command = "mix format",
+  args = { "$FILENAME" },
+}
--- a/lua/conform/formatters/mix.lua
+++ b/lua/conform/formatters/mix.lua
@@ -4,6 +4,10 @@ return {
     url = "https://hexdocs.pm/mix/main/Mix.Tasks.Format.html",
     description = "Format Elixir files using the mix format command.",
   },
-  command = "mix format",
-  args = { "$FILENAME" },
+  command = "mix",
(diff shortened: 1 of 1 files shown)


lua/conform/formatters/mix.lua
```diff
@@ -0,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://hexdocs.pm/mix/main/Mix.Tasks.Format.html",
+    description = "Format Elixir files using the mix format command.",
+  },
+  command = "mix format",
```
stevearc, line 7: I'm actually surprised this works...we do a check to see if the command is executable before formatting and this would be looking for a binary named `mix format`, instead of `mix`.

Even if this does work, I'd recommend switching it to 

```lua
command = "mix",
args = { "format", "-" }
```

I switched `$FILENAME` to `-` as well, since we want to format stdin and from the docs

> If any of the files is -, then the input is read from stdin and the output is written to stdout.

Lastly, a nice-to-have is ensuring that we run the formatter command from the correct directory (it might matter if you're editing a file in a different directory, like `../../project/file.ex`). You can do that with

```lua
cwd = require("conform.util").root_file({
  ".formatter.exs",
  "mix.exs",
})
```

stevearc: (I haven't tried any of this and don't have mix installed, so of course verify anything I wrote here and make sure it actually works)

ethangunderson: I've tested these changes and they are working locally. Interestingly, moving `format` to an arg is much slower than shoving it into the command. I had to bump my timeout from 500ms to 900ms.

Thanks for the help with this, I've pushed these changes.

stevearc: Very strange...I'm still surprised that `command = "mix format"` worked at all. But if it works now then it works `¯\_(ツ)_/¯`

stevearc (approved):
Thanks for the PR!

PR #114, opened 2023-10-05
feat: add just

pythops:
Add a formatter for [justfile](https://github.com/casey/just)


stevearc:
Thanks for the PR!

eero-lehtinen:
This config seems to be in the wrong folder so it doesn't work.

stevearc:
fixed

PR #121, opened 2023-10-06
feat: add easy-coding-standard

fbuchlak:
Add ecs php formatter ([easy-coding-standard](https://github.com/easy-coding-standard/easy-coding-standard))

### DIFF
--- a/lua/conform/formatters/easy-coding-standard.lua
+++ b/lua/conform/formatters/easy-coding-standard.lua
@@ -8,7 +8,7 @@ return {
   },
   command = util.find_executable({
     "vendor/bin/ecs",
-  }, "easy-coding-standard"),
+  }, "ecs"),
   args = { "check", "$FILENAME", "--fix", "--no-interaction" },
   cwd = util.root_file({
     "ecs.php",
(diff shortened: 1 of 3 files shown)


lua/conform/formatters/easy-coding-standard.lua
```diff
@@ -0,0 +1,18 @@
...
+
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/easy-coding-standard/easy-coding-standard",
+    description = "ecs - Use Coding Standard with 0-knowledge of PHP-CS-Fixer and PHP_CodeSniffer.",
+  },
+  command = util.find_executable({
+    "vendor/bin/ecs",
+  }, "easy-coding-standard"),
```
stevearc, line 11: Is the executable `ecs`? If so, the last argument to find_executable should be `ecs` because that's the fallback if none of the listed names are found

stevearc (changes requested):
one question

lua/conform/formatters/easy-coding-standard.lua
fbuchlak: Yes, it is. Resolved.

stevearc:
Thanks for the PR!

PR #137, opened 2023-10-13
Add ktlint formatter

shurizzle:


lua/conform/formatters/ktlint.lua
```diff
@@ -0,0 +1,10 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://ktlint.github.io/",
+    description = "An anti-bikeshedding Kotlin linter with built-in formatter.",
+  },
+  command = "ktlint",
+  args = { "--format", "--stdin", "**/*.kt", "**/*.kts" },
```
stevearc, line 8: do you need the globs? Based on [the docs](https://pinterest.github.io/ktlint/1.0.1/install/cli/#stdin-stdout) it seems like those aren't necessary if you're passing `--stdin`

lua/conform/formatters/ktlint.lua
```diff
@@ -0,0 +1,10 @@
...
+  stdin = true,
```
stevearc, line 9: this is the default and can be omitted

stevearc (changes requested):

lua/conform/formatters/ktlint.lua
shurizzle: Without paths it prints out
`18:40:35.484 [main] INFO com.pinterest.ktlint.cli.internal.KtlintCommandLine - Enable default patterns [**/*.kt, **/*.kts]`
1) I don't know how to exclude stderr from output
2) Those paths are enabled automatically

Any hints?

lua/conform/formatters/ktlint.lua
shurizzle: Usually I like to be explicit, do you want me to delete this line?

lua/conform/formatters/ktlint.lua
shurizzle: I saw null-ls.nvim and they did the same

stevearc: From the same docs:
`Logging output printed to stdout can be suppressed by setting --log-level=none`

Not sure why that _wouldn't_ be enabled by default when using stdin/stdout, but it seems like a good place to start

stevearc:
Thanks for the PR!

PR #142, opened 2023-10-14
Add blue formatter

frrenzy:
API is the same with Black formatter

### DIFF
--- a/lua/conform/formatters/blue.lua
+++ b/lua/conform/formatters/blue.lua
@@ -3,7 +3,7 @@ local util = require("conform.util")
 return {
   meta = {
     url = "https://github.com/grantjenks/blue",
-    description = "Blue -- Some folks like black but I prefer blue.",
+    description = "The slightly less uncompromising Python code formatter.",
   },
   command = "blue",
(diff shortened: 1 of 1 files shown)


lua/conform/formatters/blue.lua
```diff
@@ -0,0 +1,21 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/grantjenks/blue",
+    description = "Blue -- Some folks like black but I prefer blue.",
```
stevearc, line 6: Could you make this a bit more useful? The tagline on the repo is `The slightly less uncompromising Python code formatter.` and is more descriptive

stevearc (changes requested):

lua/conform/formatters/blue.lua
frrenzy: Sure, done! I believe this was previous tagline on the repo, but creator moved it to README :)

stevearc:
Thanks for the PR!

PR #146, opened 2023-10-17
feat: zprint formatter for clojure

agriffis:
https://github.com/kkinnear/zprint

For the config flags used here, see https://github.com/kkinnear/zprint/blob/ce5a1caf676b00f2b754e35151042211fbcc57f8/doc/reference.md?plain=1#L3727-L3745


stevearc:
Thanks for the PR!

PR #158, opened 2023-10-23
Change `rustfmt` formatter to work via `cargo`

dpecos:
Invoking `rustfmt` directly does not apply the settings set in `Cargo.toml`, like the Rust edition the file is using.

This pull request changes the command that was being used to the equivalent `cargo`, which uses `rustfmt` under the hood.

### DIFF
--- a/lua/conform/formatters/rustfmt.lua
+++ b/lua/conform/formatters/rustfmt.lua
@@ -4,6 +4,6 @@ return {
     url = "https://github.com/rust-lang/rustfmt",
     description = "A tool for formatting rust code according to style guidelines.",
   },
-  command = "rustfmt",
-  args = { "--emit=stdout" },
+  command = "cargo",
+	args = { "fmt", "--", "-q", "--emit=stdout" },
 }


TraceLTRC:
Why was this closed? I installed conform.nvim and tried to format a rust source code which has async code. With the default settings, it error'd out, saying that I need to pass `--edition 2021` to the arguments of rustfmt.

```
error[E0670]: `async fn` is not permitted in Rust 2015
             --> <stdin>:111:1
              |
          111 | async fn multipart_upload(
              | ^^^^^ to use `async fn`, switch to Rust 2018 or later
              |
              = help: pass `--edition 2021` to `rustc`
              = note: for more on editions, read https://doc.rust-lang.org/edition-guide
          
          error: `async move` blocks are only allowed in Rust 2018 or later
             --> <stdin>:134:13
              |
          134 |             async move {
              |             ^^^^^^^^^^
          
          error[E0670]: `async fn` is not permitted in Rust 2015
             --> <stdin>:189:1
              |
          189 | async fn main() -> Result<(), Box<dyn Error>> {
              | ^^^^^ to use `async fn`, switch to Rust 2018 or later
              |
              = help: pass `--edition 2021` to `rustc`
              = note: for more on editions, read https://doc.rust-lang.org/edition-guide
````

While I could pass the edition, changing the formatter to use cargo allows me to format the code without having to explicitly pass the compiler edition.

PR #159, opened 2023-10-26
fix(formatexpr): Only fallback to lsp with active clients

cryptomilk:
This fixes just formatting text in e.g. git commit messages with `gq`.


stevearc:
Oh oops, I saw your issue comment first and fixed it before seeing the PR >.<

Should be working now

PR #164, opened 2023-10-28
Fix typo in README.md

FarukhS52:


stevearc:
The README is actually autogenerated. You'll need to change the text at the source to update it
https://github.com/stevearc/conform.nvim/blob/253878436e2b6d73dfd91ccf0ac12d04cc683d34/lua/conform/formatters/rufo.lua#L4

Also the spelling for "normaliser" on HTML Beautifier is just the British version, and not incorrect. It's how they spell it on their repo, so I think it makes sense to keep it as-is.

FarukhS52:
Ohhhhh my bad,i didn't know that. Actually i am new to GitHub.
Btw thanks for the information @stevearc

PR #165, opened 2023-10-28
Add CSharpier formatter

JesperLundberg:


stevearc (changes requested):
Looks good; could you fix the luacheck error?

stevearc:
Also just a note, the documentation changes aren't necessary. You don't have to change anything, but they'll all be overwritten by the doc autogen bot after merge. For new formatters you only have to add the formatter file.

JesperLundberg:
Fixed the warning.

Also, I stole with pride from another PR here and just copied what they had done. :sweat_smile: 

Now I know til next time that I only need to add formatter file though. :)

stevearc:
No worries, I think that's how most people do it and there's no problem with it (apart from a bit more work for you). Thanks for the PR!

PR #167, opened 2023-10-30
feat: add usort

### LINKED ISSUE #156: feat: add `usort` (replace `isort`?) for python
[usort is safer than isort](https://usort.readthedocs.io/en/latest/why.html), and, imho, should replace isort, at least as a default python formatter.

fbuchlak:
Resolves #156

### DIFF
--- a/lua/conform/formatters/usort.lua
+++ b/lua/conform/formatters/usort.lua
@@ -1,0 +1,15 @@
+local util = require("conform.util")
+
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/facebook/usort",
+    description = "Safe, minimal import sorting for Python projects.",
+  },
+  command = "usort",
+  args = { "format", "$FILENAME" },
+  stdin = false,
+  cwd = util.root_file({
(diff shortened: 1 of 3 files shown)


lua/conform/formatters/usort.lua
```diff
@@ -0,0 +1,15 @@
...
+
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/facebook/usort",
+    description = "Safe, minimal import sorting for Python projects.",
+  },
+  command = "usort",
+  args = { "format", "$FILENAME" },
+  stdin = false,
```
stevearc, line 11: Could we use the syntax for [formatting via stdin](https://github.com/facebook/usort/issues/8)? It's preferable when the formatter supports it.

stevearc (changes requested):

lua/conform/formatters/usort.lua
fbuchlak: Resolved.

stevearc:
Thanks for the PR!

PR #171, opened 2023-10-30
Add markdownlint-cli2

georgeguimaraes:
markdownlint-cli2 is a better cli for markdownlint. 

It allows having config files in different directories, and it's also from the same author as the original node.js markdownlint lib.


stevearc:
Thanks for the PR!

PR #178, opened 2023-11-05
fix: filename `ast-grep`

chrisgrieser:
Sorry, overlooked the missing hyphen, until I noticed it not working here on my end. By having the filename without, adding `ast-grep` as formatter does not work, even though you'd expect it to be due to the it being spelled that way since it's the official spelling.


stevearc:
LGTM

PR #180, opened 2023-11-06
feat: add typstfmt

HakonHarnes:
Adds support for [typstfmt](https://github.com/astrale-sharp/typstfmt/).


stevearc:
Thanks for the PR!

PR #189, opened 2023-11-09
fix: remove unnecessary `resolve_formatters` call in injected formatter

### LINKED ISSUE #187: bug: injected formatter not working
### Neovim version (nvim -v)

NVIM v0.10.0-dev-1496+g4c8fdc018b

### Operating system/version

Arch Linux

### Add the debug logs

### Log file

```
Log file: /tmp/nvim-minimal/state/nvim/conform.log                                                                                                                                                             │
``│          17:29:56[INFO] Run prettierd on /home/pedro/Desktop/foo.md                                                                                                                                           │
~ │          17:29:56[DEBUG] Run command: { "prettierd", "/home/pedro/Desktop/foo.md" }                                                                                                                           │
~ │          17:29:56[DEBUG] Run CWD: /home/pedro                                                                                                                                                                 │
~ │          17:29:57[DEBUG] prettierd exited with code 0                                                                                                                                                         │
~ │          17:29:57[INFO] Run injected on /home/pedro/Desktop/foo.md                                                                                                                                            │
~ │          17:29:57[DEBUG] Injected format python:4:4: { {                                                                                                                                                      │
~ │              available = true,                                                                                                                                                                                │
~ │              command = "black",                                                                                                                                                                               │
~ │              name = "black"                                                                                                                                                                                   │
~ │            } }         

Formatters for this buffer:                                                                                                                                                                                    │
~ │prettierd ready (markdown)                                                                                                                                                                                     │
~ │injected ready (markdown)                                                                                                                                                                                      │
```                                                                                                                                                                                       

### Describe the bug

The injected formatter doesn't appear to update the injected code in a markdown file

### Steps To Reproduce

1. Using the minimal init.lua file open nvim
2. edit a foo.md file with
```md
# H1

```python
x=2
` ``
```
note there are closing backticks after the code block actually written as ``` (without the space)
3. `:w` write the file.
4. The  python code is not formatted

### Expected Behavior

Get the code block formatted as 
```python
x = 2
```

### Minimal example file

_No response_

### Minimal init.lua

```Lua
local root = '/tmp/nvim-minimal'

-- Set stdpaths to use root dir
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
    vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end

-- Bootstrap lazy
local lazypath = root .. '/plugins/lazy.nvim'
    vim.fn.system({
        'git',
        'clone',
        '--filter=blob:none',
        '--single-branch',
        'https://github.com/folke/lazy.nvim.git',
        lazypath,
    })
end

-- Install plugins
local plugins = {
    {
        'nvim-treesitter/nvim-treesitter',
        build = ':TSUpdate',
        config = function()
            require('nvim-treesitter.configs').setup({
                ensure_installed = {
                    'markdown',
                    'markdown_inline',
                    'python',
                },
            })
        end,
    },
    {
        'stevearc/conform.nvim',
        config = function()
            require('conform').setup({
                formatters_by_ft = {
                    markdown = { 'prettierd', 'injected' },
                    python = { 'black' },
                },
                format_on_save = function()
                    return { timeout_ms = 700, quiet = true, lsp_fallback = false }
                end,
                notify_on_error = true,
            })
        end,
    },
}
require('lazy').setup(plugins, {
    root = root .. '/plugins',
})
```

### Additional context

conforminjected

mehalter:
It looks like this is already being handled in the `format_lines` call. We just need the names here

Resolves #187

### DIFF
--- a/lua/conform/formatters/injected.lua
+++ b/lua/conform/formatters/injected.lua
@@ -170,12 +170,9 @@ return {
         num_format = num_format + 1
         last_start_lnum = start_lnum
         local input_lines = util.tbl_slice(lines, start_lnum, end_lnum)
-        local ft_formatters = conform.formatters_by_ft[lang]
-        local formatter_names
-        if type(ft_formatters) == "function" then
-          formatter_names = ft_formatters(ctx.buf)
-        else
-          formatter_names = require("conform").resolve_formatters(ft_formatters, ctx.buf, false)
+        local formatter_names = conform.formatters_by_ft[lang]
+        if type(formatter_names) == "function" then
+          formatter_names = formatter_names(ctx.buf)
         end
         local format_opts = { async = true, bufnr = ctx.buf, quiet = true }
         local idx = num_format


stevearc:
It's a little more subtle than this; we _do_ still need the call to `resolve`, we just need to fetch the names of the resolved formatters 68abada5a348f448eabdbd7d71884c195969484f. Thanks for looking into this!

mehalter:
ah, that was my first thought as well and opted for this when I saw the call it makes with the list calls the resolution function, but I haven't fully digested the code. Thanks for patching this and all your amazing plugins 🤗

mehalter:
Out of curiosity, could you describe the reasoning for calling the resolution on the bare list rather than just passing the list from the `formatters_by_ft`? Is this because there could be nested lists and things that need to be resolved before calling the format function?

mehalter:
Ah I totally relate on rushing through issues before meetings 😅 ! I also was curious about that type annotation as well, thanks for clarifying. Totally agree with aligning the behavior with `conform.format` being a good idea. Thanks for the insights!

PR #196, opened 2023-11-11
feat: add `auto-optional`

filipgodlewski:
A Python tool. Quite useful for lazy people like me ;)

Unfortunately, I noticed that using this tool might require user to increase the timeout. This doesn't happen in null-ls, but that's for another story.


stevearc:
Thanks for the PR!

PR #198, opened 2023-11-14
feat: add cbfmt

G0V1NDS:
Added support for [cbfmt](https://github.com/lukas-reineke/cbfmt) formatter to format codeblocks inside markdown and org documents


stevearc:
Thanks for the PR!

PR #206, opened 2023-11-17
chore: remove $FILENAME from standardjs formatter config

cesconix:
I've removed the `$FILENAME` argument from the **standardjs** formatter configuration. It's not needed with the `--stdin` flag.


stevearc:
LGTM thanks!

PR #213, opened 2023-11-21
feat/add-sqlfluff

fbuchlak:


lua/conform/formatters/sqlfluff.lua
```diff
@@ -0,0 +1,20 @@
+local util = require("conform.util")
+
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/sqlfluff/sqlfluff",
+    description = "A modular SQL linter and auto-formatter with support for multiple dialects and templated code.",
+  },
+  command = "sqlfluff",
+  args = { "fix", "--force", "--dialect=ansi", "-" },
```
stevearc, line 10: I see that the CLI docs list `sqlfluff format`
> This effectively force applies sqlfluff fix with a known subset of fairly stable rules

Could you give a brief explanation of why it makes sense to use one over the other?

lua/conform/formatters/sqlfluff.lua
```diff
@@ -0,0 +1,20 @@
...
+  stdin = true,
+  cwd = util.root_file({
+    ".sqlfluff",
+    "pep8.ini",
+    "pyproject.toml",
+    "setup.cfg",
+    "tox.ini",
```
stevearc, line 17: These all look python-specific. Is there a reason to include them as root files for sqlfluff?

stevearc (changes requested):
Couple Q's

lua/conform/formatters/sqlfluff.lua
fbuchlak: https://docs.sqlfluff.com/en/stable/configuration.html

lua/conform/formatters/sqlfluff.lua
fbuchlak: For format [documentation](https://docs.sqlfluff.com/en/stable/cli.html#sqlfluff-format) states

This effectively force applies sqlfluff fix with a known subset of fairly stable rules. **Enabled rules are ignored**, but rule exclusions (via CLI) or config are still respected.

test.sql
```sql
select foo.a, foo.b, bar.c from foo
left join bar on  bar.a = foo.a and bar.b = foo.b
```

.sqlfluff
```
[sqlfluff]
dialect=ansi

[structure.join_condition_order]
preferred_first_table_in_join_clause = earlier
```

Using format
```sql
SELECT
    foo.a,
    foo.b,
    bar.c
FROM foo
LEFT JOIN bar ON bar.a = foo.a AND bar.b = foo.b # linter error based on config file
```

Using fix
```
SELECT
    foo.a,
    foo.b,
    bar.c
FROM foo
LEFT JOIN bar ON foo.a = bar.a AND foo.b = bar.b # fixed as specified in config
```

stevearc:
Thanks for the explanation!

PR #221, opened 2023-11-27
Use the filetype rather than the filename for setting the parser

mehalter:
this also fixes the `injected` formatter to actually load the new temp buffer it creates so that things like `filetype` are set which are required by some formatters.

I looked to add this to `prettierd` but it doesn't have the option to specify a specific language parser and will only infer it from the filename

### DIFF
--- a/lua/conform/formatters/prettier.lua
+++ b/lua/conform/formatters/prettier.lua
@@ -6,10 +6,17 @@ return {
     description = [[Prettier is an opinionated code formatter. It enforces a consistent style by parsing your code and re-printing it with its own rules that take the maximum line length into account, wrapping code when necessary.]],
   },
   command = util.from_node_modules("prettier"),
-  args = { "--stdin-filepath", "$FILENAME" },
+  args = function(ctx)
+    return { "--parser=" .. vim.bo[ctx.buf].filetype }
+  end,
   range_args = function(ctx)
     local start_offset, end_offset = util.get_offsets_from_range(ctx.buf, ctx.range)
-    return { "$FILENAME", "--range-start=" .. start_offset, "--range-end=" .. end_offset }
+    return {
+      "$FILENAME",
+      "--parser=" .. vim.bo[ctx.buf].filetype,
+      "--range-start=" .. start_offset,
+      "--range-end=" .. end_offset,
+    }
   end,
   cwd = util.root_file({
     -- https://prettier.io/docs/en/configuration.html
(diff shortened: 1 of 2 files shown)


lua/conform/formatters/prettier.lua
```diff
@@ -6,10 +6,17 @@ return {
...
+  end,
   range_args = function(ctx)
     local start_offset, end_offset = util.get_offsets_from_range(ctx.buf, ctx.range)
-    return { "$FILENAME", "--range-start=" .. start_offset, "--range-end=" .. end_offset }
+    return {
+      "$FILENAME",
+      "--parser=" .. vim.bo[ctx.buf].filetype,
+      "--range-start=" .. start_offset,
+      "--range-end=" .. end_offset,
+    }
```
stevearc, line 19: What's the purpose of this change? Is the current prettier formatter not using the correct parser in some cases?

stevearc (changes requested):
One Q

lua/conform/formatters/prettier.lua
stevearc: I think we'll need to introduce some sort of filetype-to-parser map then. The main cases I see missing from [the list of parsers](https://prettier.io/docs/en/options.html#parser) are:

* javascript
* javascriptreact
* typescriptreact

stevearc: One thing I'm working on (slowly, with the limited time that I have -_-) is the ability to configure formatters with `options`. Currently the `injected` formatter supports them, but that's the only one. I have to make a breaking API change to do this, which is why it's slow going, but I think I've found a way to do it safely and _without_ breaking anything. If I manage to ship that, we could update the prettier parser to have an options table that allows you to override the parser for a specific filetype or extension. That way you could keep the default file extension logic for the common case, and when you encounter something like your `.qmd` files it's relatively straightforward to override. What do you think?

mehalter: Oh yeah that sounds perfect and a really great solution and would be a lot better for maintainability than this PR. I'll go ahead and close this with that improvement coming down the road in mind!

Also no worries about the amount of time it's taking, you are maintaining a lot of amazing plugins which naturally takes a lot of time! Let me know if you want/need any help at all!

Elias-Graf:
Hello! Was something like this implemented? I currently have the issue, that I have a file named `config` (not the missing file extensions), which contains JSON. I now tried to set the file type to JSON (`set filetype=json`), but `prettier` fails with:

```
 09:37:05[ERROR] Formatter 'prettier' error: [error] No parser could be inferred for file "<path>/config".
```

Adding this to the conform configuration fixes it, but it would be nice if that were the standard behavior:

```lua
conform.formatters.prettier = {
  prepend_args = function()
    return { '--parser', vim.bo.filetype }
  end,
}
```

PR #233, opened 2023-12-05
refactor!: formatter config functions take self as first argument

stevearc:
cc @mehalter related to #221 

With this change, we can start adding `options` tables to formatters, which will open up more customization opportunities for formatters that need it (e.g. prettier)


mehalter:
This looks very cool! I should have some time later today or tomorrow morning to take a deeper look at this.

I could also look to test this with the `prettier` case that I have. If you want, this could end up being an additional example to put in the documentation to provide support to users as to why something like this is necessary and how to use it :)

stevearc:
Feel free to build out some functionality for prettier if you like, but there's no rush. I'll merge this in the next day or so and plan to follow up with generating documentation.

wookayin:
Thanks for the compatibility layer, not completely breaking existing configs.

Issue #55, opened 2023-09-15
feat: Allow to set `formatexpr`


cryptomilk:
As null-ls has been archived, I've started to look into alternatives for formatting. According to https://github.com/neovim/neovim/pull/24338 `:help formatexpr` should be used. By default the LSP implementation sets:

```
formatexpr=v:lua.vim.lsp.formatexpr()
```

It would be nice if conform would set it to a formatter lua functions it provides. The vim.lsp.formatexpr() is here:

https://github.com/neovim/neovim/blob/35e50d79c630b05f67a840ebe21b4043ba9a6066/runtime/lua/vim/lsp.lua#L2394-L2395

It would be nice to have a similar function in conform set it for configured filetype in conform. I can look into this in the next weeks if there is interest.

stevearc:
Added. You can use it with `vim.o.formatexpr = "v:lua.require'conform'.formatexpr()"`

cryptomilk:
Thanks for implementing it so fast!

I'm using:

```lua
    local conform = require('conform')

    conform.setup()

    local pattern = {}
    if vim.fn.executable('stylua') == 1 then
        table.insert(pattern, 'lua')
        conform.formatters_by_ft.lua = { 'stylua' }
    end

    if vim.fn.executable('shfmt') == 1 then
        local cwd = vim.fn.getcwd()
        if string.find(cwd, 'samba', 1, true) then
            require('conform.formatters.shfmt').args =
                { '--posix', '--indent', '0', '-filename', '$FILENAME' }
        else
            require('conform.formatters.shfmt').args =
                { '--indent', '4', '-filename', '$FILENAME' }
        end

        table.insert(pattern, 'sh')
        conform.formatters_by_ft.sh = { 'shfmt' }
    end

    vim.api.nvim_create_autocmd('FileType', {
        pattern = pattern,
        callback = function(args)
            vim.api.nvim_buf_set_option(
                args.buf,
                'formatexpr',
                'v:lua.require("conform").formatexpr()'
            )
        end,
    })
```

This works great.

cryptomilk:
Nice, thanks. You might want do add a few of those details to the readme!

cryptomilk:
Actually, there is an issue if `vim.o.formatexpr = "v:lua.require'conform'.formatexpr()"` is set for files not providing a formatter by conform or use LSP.

I can't format text with `gq` in a normal text files.

stevearc:
[...]

PR #238, opened 2023-12-07
fix(formatexpr)!: do not fallback to the built-in formatter

wookayin:
This makes the behavior of formatexpr more consistent with
`vim.lsp.formatexpr`; do not run vim's builtin formatter.

Problem: When conform's formatexpr is used and (range or buffer) but
conform can't do formatting and there is no LSP server with the
formatting capabilities, it will fall back to the (wrong) built-in
formatting, which might result in simply concatenating all the words.

This is a breaking change (as a part of v5.0 with breaking change),
reverting the behavior introduced in 278bcd8bf2017e187e963b515017341fdd87fe2f (#55). /cc @cryptomilk

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -719,7 +719,8 @@ M.formatexpr = function(opts)
     -- No formatters were available; fall back to lsp formatter
     return vim.lsp.formatexpr({ timeout_ms = opts.timeout_ms })
   else
-    return 1
+    -- Do not fallback to built-in formatter.
+    return 0
   end
 end
 


wookayin:
If one still would like to fallback to the built-in formatter, they can write a custom formatexpr function like (although not recommended):

```lua
function _G.formatexpr()
    if #require('conform').list_formatters_for_buffer() > 0 or
       #require("conform.lsp_format").get_format_clients() > 0
        return ('conform').formatexpr()  -- uses conform or LSP
    end
    return 1  -- fallback to the built-in formatexpr, see :help formatoptions
end
```

But ideally formatexpr should be set **locally** using LspAttach, FileType autocmds, rather than globally.
```
vim.opt_local.formatexpr = 'v:lua.require("conform").formatexpr { lsp_fallback = true }'
```

stevearc:
I think you make a good point: `gw` is always available for people that want the default behavior, and not falling back to the default formatexpr makes the behavior more consistent with the LSP formatexpr, which I generally support. The more conform's API behaves like LSP, the fewer surprises users will encounter.

I would argue that this isn't really a breaking change, it's just a behavior change. Nobody's config is going to throw errors because of this.

wookayin:
Thank you for merging this!

PR #241, opened 2023-12-07
feat(prettier): add `options` for configuring prettier parser based on filetype

mehalter:
This is a work in progress for designing some nice to use options for `prettier` (and ideally `prettierd` at some point) to enable formatting using the `filetype` rather than filename. This helps when using files with non-traditional extensions (ex. `qmd` Quarto files, `rmd` R Markdown files) as well as injected code.

It also allows the user to easily configure prettier as well if a user wants to use a different javascript parser for example, prettier supports several different types just for that one file format.

The design here is definitely still up in the air and I am open to suggestions!

Potential ideas to explore:

- Having a simple list for mapping filetypes, extensions
- Having a full override function to allow the user to do arbitrary lua operations to derive the parser information (maybe this can just be "override the `args` function yourself if you want this)

### DIFF
--- a/lua/conform/formatters/prettier.lua
+++ b/lua/conform/formatters/prettier.lua
@@ -1,15 +1,62 @@
 local util = require("conform.util")
+
+--- Helper function to parse options to into a parser if available
+---@param self conform.JobFormatterConfig
+---@param ctx conform.Context|conform.RangeContext
+---@return string[]|nil args the arguments for setting a `prettier` parser if they exist in the options, nil otherwise
+local function eval_parser(self, ctx)
+  local ft = vim.bo[ctx.buf].filetype
+  local ext = vim.fn.fnamemodify(ctx.filename, ":e")
+  local options = self.options
+  local parser = options
+    and (
+      (options.ft_parsers and options.ft_parsers[ft])
+      or (options.ext_parsers and options.ext_parsers[ext])
+    )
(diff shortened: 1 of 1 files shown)


stevearc (commented):
From my current understanding, I think what makes the most sense is having a mapping of extension->parser, another mapping of filetype->parser, and if neither of those match the file, default to passing the filename and no explicit parser. I think this would eliminate the need to have all of the default values. What do you think?

Agree that there may be some need/desire to have a function to allow complete customization, but unless you can think of a case where that would be necessary I'd say overriding the `args` is the correct way to do that.

mehalter:
Thanks for the review @stevearc ! I think that makes sense. I updated the implementation to follow what you said. I have it preferring filetype detection over extension since one is configured by the editor and can be modified at runtime easily so it seems like it would be the most intentional.

I also went ahead and removed the default table of filetype mappings. The reason I added it is because it can seem a bit funky when you set up a formatter based on filetype and then if you open something of that filetype it fails to execute, but that can be left up to the user to correctly configure their formatters.

I think if you like the options table structure, then this can be ready to go! Also I left the default table commented just to show a user looking at the code what valid options are there. In the future once automatic documentation generation is implemented, I can also make sure to add the necessary comments/structure for this to be compatible!

stevearc:
Awesome, this looks good!

Issue #225, opened 2023-11-29
Feature Request: Display location of the formatters command


agoodshort:
## Description

Is there an interest to display the location of the command found and used for a formatter in `:ConformInfo`?

Let me know your thoughts, and I can help with a PR if you want.

## Example scenario

- latest version of `prettier` installed globally (or/and through [mason.nvim](https://github.com/williamboman/mason.nvim)) for some random js file editing on my machines
- specific versions of `prettier` installed as `devDependencies` in my projects

How can I confirm/know that when I am in my projects, the `devDependencies prettier` will be used before my globally installed ones?

## Request behaviour

Display the location of the formatter command used to format the buffer. For example:

![image](https://github.com/stevearc/conform.nvim/assets/33832653/3d2a1203-bcbb-4153-b782-5aed7b05dadd)

## Actual behaviour

At the moment, the only cue available about the location are if the formatter is available or not, either global or as a project `devDependencies`

### Scenario 1

Prettier installed:
- globally ❌
- mason ❌
- project `devDependencies` ❌

![image](https://github.com/stevearc/conform.nvim/assets/33832653/60a3b9ae-b4d4-4386-a987-665d836e088a)

_Side note: in this scenario, running the formatting command does not display an error message even though a formatter is unavailable. Do you want a separate issue for this?_

### Scenario 2

Prettier installed:
- globally ❌
- mason ❌
- project `devDependencies` ✔️

![image](https://github.com/stevearc/conform.nvim/assets/33832653/375c11b8-d610-4006-8317-91a10e802a69)

### Scenario 3

Prettier installed:
- globally ✔️
- mason ❌
- project `devDependencies` ✔️

![image](https://github.com/stevearc/conform.nvim/assets/33832653/ede2f37d-896f-449c-a985-884cf5313a2d)

![image](https://github.com/stevearc/conform.nvim/assets/33832653/1d23ae21-74d4-4238-9bd0-38282e9119b3)

### Scenario 4

Prettier installed:
- globally ✔️
- mason ✔️
- project `devDependencies` ✔️

![image](https://github.com/stevearc/conform.nvim/assets/33832653/27ae02f3-28ff-451d-a5eb-dc757f0844e3)

![image](https://github.com/stevearc/conform.nvim/assets/33832653/5ef06977-4184-4fa1-b600-bf68fae461d7)

## Workaround

[...]

PR #244, opened 2023-12-09
feat: add path in ConformInfo

agoodshort:
Here is my draft for #225 

2 questions:
- Should this be an option or built-in?
- Are you okay with the type `formatter.type` being a `string?` set as `nil` when initialized?

Any thoughts?

image


gegoune:
Adding extra options for such niche setting doesn't sound right. How about using conceal?

lua/conform/init.lua
```diff
@@ -3,6 +3,7 @@ local M = {}
 ---@class (exact) conform.FormatterInfo
 ---@field name string
 ---@field command string
+---@field path? string
```
stevearc, line 6: We don't really need to add the path here, do we? We can just run `exepath` from the info window where we need the full path

lua/conform/health.lua
```diff
@@ -114,12 +114,25 @@ M.show_window = function()
...
       table.insert(lines, line)
       table.insert(
         highlights,
         { "DiagnosticInfo", #lines, formatter.name:len(), formatter.name:len() + 6 }
       )
+      if formatter.path ~= nil then
+        table.insert(highlights, {
+          "DiagnosticInfo",
+          #lines,
+          formatter.name:len() + 7 + table.concat(filetypes, ", "):len() + 3,
```
stevearc, line 132: Let's pull the `table.concat(filetypes, ", ")` into a variable so we don't have to run it twice

stevearc (changes requested):
Not sure what you mean with your question about `formatter.type`
I think it's fine to include the binary path all the time. No need for an option. The info panel should only be used for debugging, so it's fine if it's a little visually busy.

Couple changes requested, but generally on board with the approach

lua/conform/init.lua
agoodshort: Sorry made a mistake, I meant `formatter.path` instead of `type` but here you're answering my question.
Yeah, I think we don't as we use it only in the info panel, so it can be produced there.

lua/conform/health.lua
agoodshort: Will do

agoodshort:
@stevearc I just squashed the requested changes into my commit.

lua/conform/health.lua
```diff
@@ -114,12 +114,22 @@ M.show_window = function()
...
-      local line = string.format("%s ready (%s)", formatter.name, table.concat(filetypes, ", "))
+      local filetypes_list = table.concat(filetypes, ", ")
+      local path = vim.fn.exepath(formatter.command)
+      local line = string.format("%s ready (%s) %s", formatter.name, filetypes_list, path)
       table.insert(lines, line)
       table.insert(
         highlights,
         { "DiagnosticInfo", #lines, formatter.name:len(), formatter.name:len() + 6 }
       )
+      if path ~= nil then
```
stevearc, line 125: `vim.fn.exepath` never returns nil. If the executable is missing, it will return an empty string

stevearc (changes requested):
One nit then LGTM!

lua/conform/health.lua
agoodshort: Oh right! Thanks for taking the time to review thoroughly

stevearc:
Great, thanks for the PR!

PR #245, opened 2023-12-10
feat: add hindent for haskell formatting

jfvillablanca:
I'm unsure what `stdin = true` means so I excluded it. Is it necessary?

### DIFF
--- a/lua/conform/formatters/hindent.lua
+++ b/lua/conform/formatters/hindent.lua
@@ -1,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://hackage.haskell.org/package/hindent",
+    description = "Extensible Haskell pretty printer.",
+  },
+  command = "hindent",
+  args = { "$FILENAME" },
+}


lua/conform/formatters/hindent.lua
```diff
@@ -0,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://hackage.haskell.org/package/hindent",
+    description = "Extensible Haskell pretty printer.",
+  },
+  command = "hindent",
+  args = { "$FILENAME" },
```
stevearc, line 8: From the docs it sounds like hindent reads stdin and writes stdout by default, and that passing a file name will format the file in-place. Conform has a mild preference for stdin/stdout formatting. Does this still work if you remove the `args` entirely?

stevearc (changes requested):

lua/conform/formatters/hindent.lua
jfvillablanca: i haven't checked and to be perfectly honest, this is my first time contributing something that involves the neovim ecosystem, so I don't have the slightest idea on how to test 😅
Can you point me in the right direction on how to do that

stevearc: Ah, no worries! There are a lot of ways to do this, but I'll give you the steps for how I typically go about it. The first thing you have to do is get your changes to `conform.nvim` to be loaded into your config. You will already have a `conform.nvim` repo somewhere on your disk, cloned by your package manager. The easy way to find it is to open vim and run `:set runtimepath?` to view all the runtime paths. There should be a path in here with `conform.nvim` in the name. That's where the repo currently is.

Go to that repo and check out your changes. I typically do this by changing the origin to point to my fork
```
git remote set-url origin git@github.com:jfvillablanca/conform.nvim
git fetch
git checkout -b hindent origin/hindent
```

Now the version of conform.nvim you're running should have your changes. You should be able to make changes in this repo, restart vim, and see the changes immediately reflected. The change I wanted you to test is just deleting the `args = ...` line entirely.

Now you just need to add the `hindent` formatter to your conform config, and test that it works when you want to format files. If it doesn't, you can always set `log_level = vim.log.levels.DEBUG` in `conform.setup` and check the log file (the location of which will be displayed in `:ConformInfo`).

stevearc:
Support added in #430

PR #246, opened 2023-12-11
Execute rubocop via rubocop's server

pjg:
This ensures a much quicker execution on 2nd and any subsequent run. Otherwise there's always a 1-3 seconds delay before formatting is being applied (as rubocop is not that quick).

From rubocop's documentation:

```
--server If a server process has not been started yet, start the server process and execute inspection with server.
```

([documentation](https://docs.rubocop.org/rubocop/usage/server.html))

I've tested it locally with custom defined formatter and it resulted in a much quicker execution on subsequent runs. Best if someone else also tested this to confirm my observations.

An even better approach would be to use the [LSP server](https://docs.rubocop.org/rubocop/usage/lsp.html), but I'm not sure how to achieve that (specifying the `--lsp` option wipes the buffer on each conform run for me).

### DIFF
--- a/lua/conform/formatters/rubocop.lua
+++ b/lua/conform/formatters/rubocop.lua
@@ -6,6 +6,7 @@ return {
   },
   command = "rubocop",
   args = {
+    "--server",
     "-a",
     "-f",
     "quiet",


pjg:
Thanks! As far as rubocop's usage via LSP is concerned, my use case is slightly more complicated as I'm using conform to run both rubyftm and rubocop, one after the other. And rubyfmt currently only on certain file patterns. Ideally I should use just one tool for the job, which allows we to run formatters/linters sequentially (which conform.nvim does). In other words, I believe conform should also allow me to run rubocop via LSP as one of the options to run it.

When using two tools, both attaching to `BufWritePre`, I'm not sure I will be able to maintain correct order of execution, something which is crucial to me. I'll experiment with it, though.

pjg:
After some trials and tribulations I was able to make it all work the way I want. First, rubyfmt is being executed by conform.nvim, and after it finishes (sync execution), the result is processed by rubocop via LSP.

References:
* conform setup: https://github.com/pjg/dotfiles/commit/2d6c577cba8ea0cd1f7e931489354b2e3097d1c0
* full file: https://github.com/pjg/dotfiles/blob/main/.vim/lua/init.lua

PR #254, opened 2023-12-27
Fix(black): formatting excluded files results in blank buffer (#249)

### LINKED ISSUE #249: bug: black formatter results in blank buffer if using `--force-exclude` option
### Neovim version (nvim -v)

0.9.4

### Operating system/version

linux

### Add the debug logs

### Log file

Log file: /conform_bug/.repro//state/nvim/conform.log
          23:56:48[INFO] Run black on /conform_bug/test.py
          23:56:48[DEBUG] Run command: { "black", "--stdin-filename", "/conform_bug/test.py", "--quiet", "-" }
          23:56:48[DEBUG] Run CWD: /conform_bug
          23:56:48[DEBUG] black exited with code 0

### Describe the bug

`black` can be configured to ignore files that match a regex pattern with the `--force-exclude` option. Matching files will be ignored by black and not formatted. A python project can add this config option to `pyproject.toml` file using `tools.black.force-exclude`

Running `conform.format()` on a file ignored by `black` results in an empty buffer.

blocking (cannot use plugin)

### Steps To Reproduce

Here's a gist: https://gist.github.com/bartels/c71e95c5d73f96952c184567ac4a19cd

### Expected Behavior

The buffer should not be modified since black is set to ignore them.

### Minimal example file

test.py

```python
#!/usr/bin/env python

# This file is ignored by black with `force-exclude` option.
# Any files contents will do. conform.format() will result in a blank buffer

print("No, please don't delete me!!!")
```
pyproject.toml

```toml
# Ignore test.py
[tool.black]
force-exclude = '''
^/test\.py
'''
```

### Minimal init.lua

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/conform.nvim",
    config = function()
      require("conform").setup({
        formatters_by_ft = {
          python = { 'black' },
        }
      })
    end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

### Additional context

When you run `black` on an ignored file, it will not output the file contents, but  it does  have a return code of `0`.

```sh
cat test.py | black --quiet --stdin-filename "$PWD"/test.py -
echo $?
```

The empty output is likely the problem. Can conform handle this case?  Is there a way to configure conform to ignore empty output from formatters?

bartels:
Fixes #249 

This fix checks for blank formatter output in `apply_format`. Not sure if this is the best place to apply this fix, but seemed  the most obvious place since it'll work the same for both sync and async formats

### DIFF
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -172,7 +172,7 @@ M.apply_format = function(bufnr, original_lines, new_lines, range, only_apply_ra
   -- Abort if output is empty but input is not (i.e. has some non-whitespace characters).
   -- This is to hack around oddly behaving formatters (e.g black outputs nothing for excluded files).
   if new_text:match("^%s*$") and not original_text:match("^%s*$") then
-    log.trace("Aborting because a formatter returned empty output")
+    log.warn("Aborting because a formatter returned empty output for buffer %s", bufname)
     return
   end
 
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -171,7 +171,7 @@ M.apply_format = function(bufnr, original_lines, new_lines, range, only_apply_ra
 
   -- Abort if output is empty but input is not (i.e. has some non-whitespace characters).
   -- This is to hack around oddly behaving formatters (e.g black outputs nothing for excluded files).
-  if new_text:match("^%s*$") and not original_text:match('^%s*$') then
+  if new_text:match("^%s*$") and not original_text:match("^%s*$") then
     log.trace("Aborting because a formatter returned empty output")
     return
   end
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -169,6 +169,13 @@ M.apply_format = function(bufnr, original_lines, new_lines, range, only_apply_ra
   table.remove(original_lines)
   table.remove(new_lines)
 
+  -- Abort if output is empty but input is not (i.e. has some non-whitespace characters).
+  -- This is to hack around oddly behaving formatters (e.g black outputs nothing for excluded files).
+  if new_text:match("^%s*$") and not original_text:match('^%s*$') then
+    log.trace("Aborting because a formatter returned empty output")
+    return
+  end
+
   log.trace("Comparing lines %s and %s", original_lines, new_lines)
   local indices = vim.diff(original_text, new_text, {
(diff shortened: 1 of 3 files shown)


lua/conform/runner.lua
```diff
@@ -169,6 +169,13 @@ M.apply_format = function(bufnr, original_lines, new_lines, range, only_apply_ra
   table.remove(original_lines)
   table.remove(new_lines)
 
+  -- Abort if output is empty but input is not (i.e. has some non-whitespace characters).
+  -- This is to hack around oddly behaving formatters (e.g black outputs nothing for excluded files).
+  if new_text:match("^%s*$") and not original_text:match("^%s*$") then
+    log.trace("Aborting because a formatter returned empty output")
+    return
+  end
```
stevearc, line 177: I think this is a fine place for it, but I'd say add the buffer name to the log line and make it a `warn` instead of a `trace`

tests/runner_spec.lua
```diff
@@ -272,6 +272,19 @@ print("a")
...
     end)
 
+    it("discards formatting changes if formatter output is empty /w non-empty input", function()
+      local bufnr = vim.fn.bufadd("testfile")
+      vim.fn.bufload(bufnr)
+      vim.api.nvim_set_current_buf(bufnr)
+      local original_lines = { "line one", "line two" }
+      vim.api.nvim_buf_set_lines(bufnr, 0, -1, true, original_lines)
+      vim.bo[bufnr].modified = false
+      test_util.set_formatter_output({ "" })
```
stevearc, line 282: I recently made a change the moved this out of `test_util`. You may need to rework the test a bit.

stevearc (changes requested):
Approach looks good, but the fuzz tester is failing. You may need to adjust the fuzz tester to not trigger this condition

stevearc:
Thanks for the PR!

Mange:
Thanks! This helped me with the `rubocop` formatter that also started clearing all my files on format for no discernible reason after updating 2 days ago.

I'll try to figure out why *that* happens and report it later on, but now I can get back to work. :heart:

PR #257, opened 2023-12-31
feat: add definition for LuaFormatter

Roguelazer:
This adds [LuaFormatter](https://github.com/Koihik/LuaFormatter) as a formatter for lua. As far as I can tell, it doesn't support stdin/stdout (passing an argument of `-` attempts to format a literal file named `-`). The `-i` argument makes it apply changes in-place instead of returning them on stdout.

### DIFF
--- a/lua/conform/formatters/luaformatter.lua
+++ b/lua/conform/formatters/luaformatter.lua
@@ -1,0 +1,13 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/Koihik/LuaFormatter",
+    description = "Reformats your Lua source code.",
+  },
+  command = "lua-format",
+  args = { "-i", "$FILENAME" },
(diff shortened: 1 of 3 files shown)


lua/conform/formatters/luaformatter.lua
```diff
@@ -0,0 +1,13 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/Koihik/LuaFormatter",
+    description = "Reformats your Lua source code.",
+  },
+  command = "lua-format",
+  args = { "-i", "$FILENAME" },
+  stdin = false,
```
stevearc, line 9: One quick check: does it format if you use `stdin = true` and don't pass any args?

stevearc (changes requested):

stevearc:
Support added in #432

PR #259, opened 2024-01-02
feat: add xmllint formatter

chefe:
Add a config for [xmllint](http://xmlsoft.org/xmllint.html) inspired by [none-ls](https://github.com/nvimtools/none-ls.nvim). 

Source: https://github.com/nvimtools/none-ls.nvim/blob/main/lua/null-ls/builtins/formatting/xmllint.lua


stevearc:
Thanks for the PR!

PR #263, opened 2024-01-04
fix: set a cwd for dprint

azzamsa:
The `dprint` cwd according to the docs:

> After [installing](https://dprint.dev/install), the main part of getting setup is to create a dprint.json/dprint.jsonc, or hidden .dprint.json/.dprint.jsonc file in your project.
> https://dprint.dev/setup/

I can confirm it has been tested on my local machine.


stevearc:
LGTM thanks!

PR #264, opened 2024-01-04
fix: set a cwd for djlint

azzamsa:
According to the docs:

> Configuration is done either through your projects pyproject.toml file, or a .djlintrc file.
> https://www.djlint.com/docs/configuration/


stevearc:
LGTM thanks!

Issue #271, opened 2024-01-08
Feature Request: Support buffer formatting when formatting the whole file


ndavd:
As it was discussed in #269, it would be helpful if conform's formatexpr took into account buffer formatting. That'd be particularly useful for when an LSP doesn't the have `vim.lsp.protocol.Methods.textDocument_rangeFormatting` capabilities and only supports `vim.lsp.protocol.Methods.textDocument_formatting`. So conform's formatexpr could check if the user is formatting the whole file and use buffer formatting in that case regardless if range formatting is supported.

(PS: Sorry for not using the proper feature request template, I'm opening this from my phone and the GitHub app was redirecting me to the browser)

PR #272, opened 2024-01-08
feat: support lsp buffer formatting when formatting the whole file

ndavd:
solves: https://github.com/stevearc/conform.nvim/issues/271

Turned out to be such a simple change... `opts.range` should just be `nil` when the whole buffer is selected.

cc @wookayin

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -721,10 +721,17 @@ M.formatexpr = function(opts)
   end
   local end_line = vim.fn.getline(end_lnum)
   local end_col = end_line:len()
-  opts.range = {
-    start = { start_lnum, 0 },
-    ["end"] = { end_lnum, end_col },
-  }
+
+  if vim.v.count == vim.fn.line("$") then
+    -- Whole buffer is selected; use buffer formatting
+    opts.range = nil
+  else
+    opts.range = {
+      start = { start_lnum, 0 },
+      ["end"] = { end_lnum, end_col },
+    }
+  end
+
   if M.format(opts) then
     return 0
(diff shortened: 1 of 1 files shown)


lua/conform/init.lua
```diff
@@ -721,10 +721,17 @@ M.formatexpr = function(opts)
   end
   local end_line = vim.fn.getline(end_lnum)
   local end_col = end_line:len()
-  opts.range = {
-    start = { start_lnum, 0 },
-    ["end"] = { end_lnum, end_col },
-  }
+
+  if vim.v.count == vim.fn.line("$") then
```
wookayin, line 725: Shouldn't `vim.v.lnum` (`start_lnum`) also be checked?

ndavd: I don't think so because it's already checking whether the number of lines to be formatted (`vim.v.count`) is equal to the total number of lines (`vimv.fn.line("$")`). Thus it must be formatting the whole buffer, no?

wookayin: I see, you're right. The count is not the last line but the length of the range, so it's only when the range represents the whole lines.

ndavd: So do you think it's good? Solves my issue and doesn't seem to break anything for me.

stevearc:
This LGTM. Ideally we would add magic support for range formatting for the LSPs that don't do it natively like we do for normal formatters, but that would be a lot of work (which is why I haven't done it yet). This is a reasonable workaround for this use case.

ndavd:
Thanks! I see, that indeed sounds great.

Issue #262, opened 2024-01-03
feature request: dry run formatters


bpjordan:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

Provide a way to "dry run" formatters, i.e. run formatters and retrieve diffs from them, but just return them to the user without actually applying them to the buffer.

### Provide background

I have a set up to format on write, but for work I often make small changes to large files that aren't formatted. I would like to be able to check if a file is formatted in a `BufOpen` autocmd and disable the autoformatting if it isn't already formatted. That way I can make the change without formatting the whole file, or manually enable formatting if I want to.

### What is the significance of this feature?

nice to have

### Additional details

I would be happy to make a PR to implement this.

My proposed API changes would be as follows:

- add a `dry_run` boolean to the `opts` table for `conform.format()`
- skip the call to `apply_format` when it is enabled
- add a second return value to `conform.format` with a table of the diffs
- add the same table as a second argument to `callback`

So my use case mentioned above could be accomplished with something like:

```lua
vim.api.nvim_create_autocmd('BufRead', {
pattern = "*",
callback = function(e)
    require('conform').format({
        dry_run = true,
        async = true,
        callback = function(err, formatter_results)
            for _, result in pairs(formatter_results) do
                if result.diff ~= nil then
                    vim.b[e.bufnr].disable_autoformat = true
                    return
                end
            end
        end}
end})
```

stevearc:
This is a neat idea, and if it can be done with relatively minimal API changes that are pretty clean, I'm on board. The API that makes the most sense to me would be:

- Add a second parameter to the `callback` function that is a boolean. This indicates if any changes were applied.
- Add a `dry_run` option to format() that will skip applying changes, but keep the same logic for that `formatted` parameter.

This way we don't expose more information than is actually needed, we neatly sidestep the issue of "how do we combine normal formatter diffs with LSP format diffs", and it's something that could be useful for non-dry_runs as well. What do you think?

bpjordan:
> This is a neat idea, and if it can be done with relatively minimal API changes that are pretty clean, I'm on board. The API that makes the most sense to me would be:
> 
>     * Add a second parameter to the `callback` function that is a boolean. This indicates if any changes were applied.
> 
>     * Add a `dry_run` option to format() that will skip applying changes, but keep the same logic for that `formatted` parameter.
> 
> 
[...]

PR #273, opened 2024-01-12
feat: Add dry_run option and report whether buffer was/would be changed by formatters

bpjordan:
Implements #262

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -362,7 +362,7 @@ end
 ---    id nil|integer Passed to |vim.lsp.buf.format| when lsp_fallback = true
 ---    name nil|string Passed to |vim.lsp.buf.format| when lsp_fallback = true
 ---    filter nil|fun(client: table): boolean Passed to |vim.lsp.buf.format| when lsp_fallback = true
----@param callback? fun(err: nil|string, changed: nil|boolean) Called once formatting has completed
+---@param callback? fun(err: nil|string, did_edit: nil|boolean) Called once formatting has completed
 ---@return boolean True if any formatters were attempted
 ---@return boolean? True if changes were made (unless async = true)
 M.format = function(opts, callback)
@@ -382,7 +382,7 @@ M.format = function(opts, callback)
   if not opts.range and mode == "v" or mode == "V" then
     opts.range = range_from_selection(opts.bufnr, mode)
   end
-  callback = callback or function(_err, _changed) end
+  callback = callback or function(_err, _did_edit) end
   local errors = require("conform.errors")
   local log = require("conform.log")
   local lsp_format = require("conform.lsp_format")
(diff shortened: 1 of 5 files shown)


lua/conform/init.lua
```diff
@@ -354,21 +354,24 @@ end
...
 ---    lsp_fallback nil|boolean|"always" Attempt LSP formatting if no formatters are available. Defaults to false. If "always", will attempt LSP formatting even if formatters are available.
 ---    quiet nil|boolean Don't show any notifications for warnings or failures. Defaults to false.
 ---    range nil|table Range to format. Table must contain `start` and `end` keys with {row, col} tuples using (1,0) indexing. Defaults to current selection in visual mode
 ---    id nil|integer Passed to |vim.lsp.buf.format| when lsp_fallback = true
 ---    name nil|string Passed to |vim.lsp.buf.format| when lsp_fallback = true
 ---    filter nil|fun(client: table): boolean Passed to |vim.lsp.buf.format| when lsp_fallback = true
----@param callback? fun(err: nil|string) Called once formatting has completed
+---@param callback? fun(err: nil|string, did_edit: nil|boolean) Called once formatting has completed
 ---@return boolean True if any formatters were attempted
+---@return boolean? True if changes were made (unless async = true)
```
stevearc, line 367: Let's not add this return value in this PR. We can discuss something like this as a follow up, but there's a more involved design decision here. I generally don't like functions that return different values based on the parameters they are called with, which is why the `err` is currently only accessible via the callback _even if_ `format()` was called synchronously. We should handle `did_edit` the same way we are handling `err`.

lua/conform/init.lua
```diff
@@ -354,21 +354,24 @@ end
 ---    timeout_ms nil|integer Time in milliseconds to block for formatting. Defaults to 1000. No effect if async = true.
 ---    bufnr nil|integer Format this buffer (default 0)
 ---    async nil|boolean If true the method won't block. Defaults to false. If the buffer is modified before the formatter completes, the formatting will be discarded.
+---    dry_run nil|boolean If true don't edit the buffer, only return whether the buffer would be edited.
```
stevearc, line 357: This is a bit deceptive, both because the `did_edit` boolean is available even when this is false and because (per the comment below) we shouldn't actually _return_ `did_edit`. I'd suggest leaving this as
> If true, don't apply formatting changes to the buffer

lua/conform/runner.lua
```diff
@@ -152,7 +152,9 @@ end
 ---@param new_lines string[]
 ---@param range? conform.Range
 ---@param only_apply_range boolean
-M.apply_format = function(bufnr, original_lines, new_lines, range, only_apply_range)
+---@param dry_run boolean
+---@return boolean?
```
stevearc, line 156: Let's change the type to be non-nilable and add a name to the return value like so:
```lua
---@return boolean any_changes
```

stevearc (changes requested):
The shape looks good! Left some comments about a few refactor changes.

stevearc:
Looks good!

PR #281, opened 2024-01-26
feat: provide diff-rename by exposing LSP TextEdits conversion functions

jedrzejboczar:
Python LSP (pylsp) resolves rename as a single TextEdit of the whole buffer. I found your diff feature of conform.nvim and wanted to adopt it for `textDocument/rename`. This only includes small changes to conform.nvim and exposes some functions that can be used to easily wrap the LSP rename handler, e.g.
```lua
local diff_rename = require('conform.lsp_format').wrap_rename_handler(vim.lsp.handlers["textDocument/rename"])
vim.lsp.handlers["textDocument/rename"] = diff_rename

-- In my config I actually have something like
vim.lsp.handlers["textDocument/rename"] = function(err, result, ctx, config)
    diff_rename(err, result, ctx, config)
    if result then 
        fill_quickfix_list_and_briefly_highlight_all_changes(result)
    end
end
```

Not sure if you want to include it in conform.nvim as it is not strictly about formatting, but if not then I'll just copy the code to my config an use it there. Still it seems like this is a small change to what conform.nvim already does. And maybe it could be useful for other LSP actions that apply text edits?

### DIFF
--- a/lua/conform/lsp_format.lua
+++ b/lua/conform/lsp_format.lua
@@ -1,38 +1,16 @@
 ---This module replaces the default vim.lsp.buf.format() so that we can inject our own logic
 local log = require("conform.log")
-local util = require("vim.lsp.util")
 
 local M = {}
 
 local function apply_text_edits(text_edits, bufnr, offset_encoding, dry_run)
-  if
-    #text_edits == 1
-    and text_edits[1].range.start.line == 0
-    and text_edits[1].range.start.character == 0
-    and text_edits[1].range["end"].line >= vim.api.nvim_buf_line_count(bufnr)
-    and text_edits[1].range["end"].character == 0
-  then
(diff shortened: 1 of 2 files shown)


jedrzejboczar:
I fixed stylua, as for typecheck maybe you could include the types from `/usr/share/nvim/runtime/lua/vim/lsp/_meta/protocol.lua` when type-checking in CI?

stevearc (changes requested):
I would like to keep conform narrowly focused on formatting. This seems to mostly be an effort to re-use the logic that diffs two blocks of text and produces LSP TextEdit objects. I _would_ be receptive to two ideas around that:
1. upstreaming the logic into Neovim itself, where it could be used to fix the behavior of all of these formatting/workspace operations
2. _light_ refactoring to make it easier to call that function as a library

PR #282, opened 2024-01-27
fix: set a cwd for biome

azzamsa:
According to the docs:

> The Biome configuration file is named `biome.json` and should be placed in the root directory of your project. The root directory is usually the directory containing your project’s `package.json`.

https://biomejs.dev/guides/how-biome-works/#configuration


stevearc:
Thanks for the PR!

Issue #211, opened 2023-11-21
bug: swift_format doesn't respect .swift-format file


heygarrett:
### Neovim version (nvim -v)

NVIM v0.9.4

### Operating system/version

macOS 14.1.1 (23B81)

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

```
Log file: /Users/garrett/dev/scrap/conform-swift/.repro//state/nvim/conform.log
          22:37:11[DEBUG] Run command: { "swift-format" }
          22:37:11[DEBUG] swift_format exited with code 0
          22:38:05[DEBUG] Running formatters on /Users/garrett/dev/scrap/conform-swift/Package.swift: { "swift_format" }
          22:38:05[INFO] Run swift_format on /Users/garrett/dev/scrap/conform-swift/Package.swift
          22:38:05[DEBUG] Run command: { "swift-format" }
          22:38:05[DEBUG] swift_format exited with code 0
          22:39:40[DEBUG] Running formatters on /Users/garrett/dev/scrap/conform-swift/Package.swift: { "swift_format" }
          22:39:40[INFO] Run swift_format on /Users/garrett/dev/scrap/conform-swift/Package.swift
          22:39:40[DEBUG] Run command: { "swift-format", "--configuration", ".swift-format" }
          22:39:40[DEBUG] swift_format exited with code 0
          22:40:05[DEBUG] Running formatters on /Users/garrett/dev/scrap/conform-swift/Package.swift: { "swift_format" }
          22:40:05[INFO] Run swift_format on /Users/garrett/dev/scrap/conform-swift/Package.swift
          22:40:05[DEBUG] Run command: { "swift-format" }
          22:40:05[DEBUG] swift_format exited with code 0

Formatters for this buffer:
swift_format ready (swift)

Other formatters:

```

### Describe the bug

When using the configuration of swift-format included in conform.nvim (listed as swift_format), the presence of a .swift-format file is not respected when formatting Swift code. Running `swift-format` from the command line _does_ respect the .swift-format file by default.

Adding `prepend_args = { "--configuration", ".swift-format" }` to the swift_format configuration does resolve the issue, but doesn't explain why the configuration file isn't respected by default.

### Steps To Reproduce

1. `mkdir conform-swift && cd conform-swift`
2. `swift package init --type executable`
3. `echo '{ "indentation": { "tabs": 1 } }' > .swift-format`
[...]

PR #283, opened 2024-01-28
fix: `swift_format` doesn't respect `.swift-format` file

NSHaoSong:
`swift-format` won't pick up `.swift-format` config file in stdin mode. We need to set `stdin = false` and `--in-place` to make it work with `conform.nvim`.

[apple/swift-format/README.md](https://github.com/apple/swift-format/blob/e312ede68bd8549381f9e2edfef7e6c176e9a50e/README.md#L129-L131)

> *   `-i/--in-place`: Overwrites the input files when formatting instead of
    printing the results to standard output. _No backup of the original file is
    made before it is overwritten._

Fix https://github.com/stevearc/conform.nvim/issues/211

### DIFF
--- a/lua/conform/formatters/swift_format.lua
+++ b/lua/conform/formatters/swift_format.lua
@@ -5,4 +5,6 @@ return {
     description = "Swift formatter from apple. Requires building from source with `swift build`.",
   },
   command = "swift-format",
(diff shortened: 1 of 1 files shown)


stevearc:
LGTM

PR #286, opened 2024-02-01
feat: add awk formatter

trevarj:
Add formatter for awk programs. Surprise! It's awk!

### DIFF
--- a/lua/conform/formatters/awk.lua
+++ b/lua/conform/formatters/awk.lua
@@ -1,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://www.gnu.org/software/gawk/manual/gawk.html",
+    description = "Format awk programs with awk",
+  },
+  command = "awk",
+  args = { "-f", "-", "-o-" },
+}


stevearc:
Love it 😆

Givo29:
Hi guys,

I know this has been closed, but would it be possible to add an additional formatter for awk, likely being [prettier-plugin-awk](https://github.com/Beaglefoot/prettier-plugin-awk/)?

Two main reasons why I'm asking this.
1. Not everyone will be using gawk (gnu awk), other versions (like mawk) come installed as the default on a lot of systems such as Debian and Ubuntu, it would be nice for there to be an option for these systems without installing gawk
2. Gawk pretty printing (as you have here) uses [profiling](https://www.gnu.org/software/gawk/manual/gawk.html#Profiling) behind the scenes, the output of which can differ a lot from what you're trying to write.. Especially in the case where you're using `@include` to import functions from another awk file, which gawk pretty-print does not like at all. prettier-plugin-awk does a much better job of this. (I can provide examples if you like).

I hope this is considered as I believe that prettier-plugin-awk is a better formatter than using gawk's pretty-print feature. It's more universal across all awk versions and also does a better job in general.

Regards,
Joel

trevarj:
@Givo29 maybe you can just add a separate "awk-prettier" formatter. Seems better since people may not be using prettier and have this extension installed.

Givo29:
I have done that for now @trevarj, I just thought I'd also make the suggestion here.

You make a valid point though, not everyone will have prettier installed either.

What I've done:
```lua
formatters = {
    awk_prettier = {
        command = "~/.nvm/versions/node/v20.17.0/bin/prettier",
        args = { "--stdin-filepath", "$FILENAME" }
    }
},
```

The only unfortunate part is I have to globally install prettier and prettier-plugin-awk as it won't work using the mason installed prettier instance...

PR #287, opened 2024-02-01
feat: add biome-check formatter

andykog:
`biome` formatter uses `biome format` command which applies formatting but doesn't apply lint autofixes. Couldn't find a better way then to create another formatter that uses `biome check --apply-unsafe`


stevearc:
Given that this is a different subcommand entirely, I think it makes sense to have it as a separate formatter. Thanks for the PR!

PR #290, opened 2024-02-04
feat(formatters): add eslint formatter

tokiory:
Added ESLint, because not all people use eslint_d.

In this case, the arguments that must be passed to eslint are exactly the same as in eslint_d, so there should be no problems.


tokiory:
Unfortunately, ESLint does not have a flag that can be used to output files to stdout. After some digging, I also realized that it is impossible to use the --fix flag with --stdin, only --fix-dry-run is available there

PR #293, opened 2024-02-07
feat: add ReScript formatter

angelofallars:
Adds the [ReScript](https://rescript-lang.org/) formatter


stevearc:
Thanks for the PR!

Issue #61, opened 2023-09-17
Explicit LSP formatter


Quaqqer:
Hi,

Great plugin. I think it would be nice to have a formatter named 'lsp' that would simply run the lsp formatter. Then it would be possible to be explicit about which filetypes would use the lsp formatter and the fallback option would not be needed anymore.

Quaqqer:
Found a closed issue about topic

stevearc:
I mentioned this at the end of #36, but I'll reiterate a bit and provide some more justification.

I explicitly want to keep the configuration of this plugin quite simple. It's very easy to fall into the trap of trying to support every possible workflow via the config, and that way leads to option hell. It can become super flexible, but it also gets very difficult to reason about, becomes harder for new users to grok, and creates almost a mini-DSL that I have to maintain, debug, and add features to. For this reason, the stock configuration is intentionally designed to cover only the most common use cases. For anything more complex, it's expected that you can call `conform.format()` directly yourself, passing in whatever list of formatters you want.

To more directly answer your immediate question: I agree that having a "lsp" formatter would be a nice interface and make the config easier. Well...in the common case it's probably much the same, but I agree that it feels a bit more aesthetic. However, the issue is that LSP formatting is decidedly _not_ like other formatting. Non-LSP formatters can operate as a pure function, taking lines in and spitting lines out. They're beautifully stateless. LSP formatters can _only_ operate on the current state of a buffer, and they return a set of TextEdits. Building a system that can treat these two the same would require additional layers of abstraction and some kludgy hacks. If all we're getting for that is a slightly nicer config interface, I don't think it's worth it. It feels better to leave things simpler and easier to reason about.

For these reasons, I'm filing this under "not planned"

Issue #36, opened 2023-09-12
using lsp formatting on specific filetypes


perrin4869:
I am trying to use lsp formatting on scala files, but avoid using lsp formatting on `typescript` files.
I have the following config:

```lua
local utils = require("utils")

require("conform").setup({
	formatters_by_ft = {
		lua = { "stylua" },
		javascript = { "eslint_d" },
		javascriptreact = { "eslint_d" },
		typescript = { "eslint_d" },
		typescriptreact = { "eslint_d" },
		-- Use a sub-list to run only the first available formatter
		json = { { "prettierd", "prettier" } },
		scala = {},
		["*"] = { "trim_whitespace" },
	},
})

local function format()
	require("conform").format({ lsp_fallback = true })
end

local function format_write()
	require("conform").format({
		lsp_fallback = true,
	}, vim.cmd.write)
end

local opts = { silent = true }
local get_opts = utils.create_get_opts(opts)
vim.keymap.set("n", "<leader>f", format, get_opts({ desc = "conform.format" }))
vim.keymap.set("n", "<leader>F", format_write, get_opts({ desc = "conform.format_write" }))

vim.api.nvim_create_autocmd("BufWritePost", {
	group = vim.api.nvim_create_augroup("FormatAutogroup", {}),
	callback = format_write,
})
```

If I run `format` on a `typescript` file with this config, `eslint_d` will format the file. On a scala file, nothing will happen.
If I replace `lsp_fallback = true -> "always"`, running `format` will run the lsp formatter on both scala and typescript.
There doesn't seem to be a way to specify running the lsp formatter on scala files, but not on typescript files?

stevearc:
Here's my recommendation for changes:
```lua
local utils = require("utils")

local function get_lsp_fallback(bufnr)
  local always_use_lsp = not vim.bo[bufnr].filetype:match("^typescript")
  return always_use_lsp and "always" or true
end

local function format()
  require("conform").format({
    lsp_fallback = get_lsp_fallback(0),
    -- I would recommend setting async=true for this so formatting doesn't freeze the editor.
    -- Async formatting changes will be discarded if you modify the buffer before it's completed.
    -- async = true,
  })
end

-- I don't personally see the need for this; if you autoformat on save, this should be the same as
-- saving the buffer. But if you like it, keep it.
local function format_write()
  require("conform").format({
    lsp_fallback = true,
    -- same comment about setting async = true here
    -- async = true,
  }, function(err)
    if not err then
[...]

PR #301, opened 2024-02-16
feat(formatters): defined a 'execute' option so a custom function can be used as a formatter

brunotvs:
An attempt to solve issues #36 and #61.


stevearc:
While I appreciate the effort that went into this, it does not solve any of the fundamental problems that are mentioned in the attached issues. LSP formatting operates on the current state of a buffer, so if you try to run this in a sequence with other formatters, it won't work properly. 

Also, there is already support for formatters that are defined via a function. That's what is used for the `injected` formatter https://github.com/stevearc/conform.nvim/blob/61cff430c9f15770d0c5e68c1b08067223bd94ab/lua/conform/formatters/injected.lua#L125

PR #312, opened 2024-02-29
feat(formatter): liquidsoap-prettier

elythh:
A formatter for https://www.liquidsoap.info/ 😄

### DIFF
--- a/lua/conform/formatters/liquidsoap-prettier.lua
+++ b/lua/conform/formatters/liquidsoap-prettier.lua
@@ -1,0 +1,11 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/savonet/liquidsoap-prettier",
+    description = "A binary to format Liquidsoap scripts",
+  },
+  command = util.from_node_modules("liquidsoap-prettier"),
+  args = { "-w", "$FILENAME" },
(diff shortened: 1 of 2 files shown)


elythh:
I think I should rather configure prettier to use that plugin.

lua/conform/formatters/liquidsoap-prettier.lua
```diff
@@ -0,0 +1,10 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/savonet/liquidsoap-prettier",
+    description = "A binary to format Liquidsoap scripts",
+  },
+  command = util.from_node_modules("liquidsoap-prettier"),
+  args = { "$FILENAME" },
```
stevearc, line 9: This command is writing the formatting to stdout, but it's reading it from the file. I think you need to set
```lua
args = { "-w", "$FILENAME" },
stdin = false,
```

stevearc (changes requested):

lua/conform/formatters/liquidsoap-prettier.lua
elythh: Both work but that way I cleaner ye

stevearc:
Thanks for the PR!

PR #313, opened 2024-02-29
feat: add OpenTofu formatter

zidhuss:
Adds `tofu_fmt` from [OpenTofu](https://opentofu.org). An alternative to `terraform_fmt` for those that have switched away from Terraform.


stevearc:
Thanks for the PR!

Issue #268, opened 2024-01-06
bug: formatter running when there is no config for the formatter in the project I'm in


NullVoxPopuli:
### Neovim version (nvim -v)

NVIM v0.10.0-dev-2042+g28c183b38

### Operating system/version

Ubuntu 23.10

### Add the debug logs

- [ ] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

I don't know how to set log level

### Describe the bug

My repo doesn't have a prettier config, yet prettier is still running.

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

My config:
```lua
  use {
    'stevearc/conform.nvim',
    config = function()
      require('conform').setup({
        formatters_by_ft = {
          -- Use a sub-list to run only the first available formatter
          javascript = { { "prettierd", "prettier" } },
          ['javascript.glimmer'] = { { "prettierd", "prettier" } },
        },
        format_on_save = {
          timeout_ms = 250,
          lsp_fallback = true,
        },
        notify_on_error = false,
      })
    end
```

### Expected Behavior

When there is no prettier config in the repo, prettier should not run

### Minimal example file

- clone https://github.com/starbeamjs/dev-compile
- have the above conform config
- have prettierd installed
- save any file, see that changes occur

### Minimal init.lua

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/conform.nvim",
    config = function()
      require("conform").setup({
        log_level = vim.log.levels.DEBUG,
        -- add your config here
      })
    end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
[...]

PR #315, opened 2024-02-29
prettierd: presence of package.json is insufficient for deciding if prettierd should attach

NullVoxPopuli:
May help with: https://github.com/stevearc/conform.nvim/issues/268

the only question we can ask if prettier should be enable is if there is a prettier config file


stevearc:
Unfortunately I can't take this change because prettier _can_ use `package.json` for its configuration (see the link in the comment https://prettier.io/docs/en/configuration.html). You could make this change only in your local config, or you could try only installing prettier locally. If the prettier executable is not found, it won't be run.

NullVoxPopuli:
silly package.json configs. thanks!

PR #321, opened 2024-03-05
Allow rust async code to be formatted by rustfmt

### LINKED ISSUE #320: bug: rust async code formatting error
### Neovim version (nvim -v)

v0.10.0-dev-2021+gee2127363

### Operating system/version

macos 14

### Add the debug logs

### Log file

image

### Describe the bug

rust async code formatting errors when using `rustfmt`;
shouldn't we add the parameter `--edition 2021`?

minor (annoyance)

### Steps To Reproduce

1. nvim main.rs
2. format the async code

### Expected Behavior

`rustfmt` support async code

### Minimal example file

_No response_

### Minimal init.lua

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/conform.nvim",
    config = function()
      require("conform").setup({
        -- add your config here
      })
    end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

### Additional context

_No response_

riftsin:
Hello,

This allows rust async code to be formatted with rustfmt.

closes #320

### DIFF
--- a/lua/conform/formatters/rustfmt.lua
+++ b/lua/conform/formatters/rustfmt.lua
@@ -5,5 +5,5 @@ return {
     description = "A tool for formatting rust code according to style guidelines.",
   },
   command = "rustfmt",
-  args = { "--emit=stdout" },
+  args = { "--emit=stdout", "--edition=2021" },
 }


TraceLTRC:
Rather than passing in the edition. It would be better to use cargo to format it.

#158 already solves this, but for some reason the author of the PR closed it before it could get discussed.

riftsin:
Hmm I'm not sure if that's easily feasible.

One of the issues with using cargo fmt instead of rustfmt is that cargo fmt requires a Cargo.toml in the working directory or a parent directory, rustfmt can be run from anywhere.

What if you are not in a rust project but still opened a rust file ?

There is an option --manifest-path that we can use to specify the path to the Cargo.toml file but can we reliably find it regardless of the current directory ?

stevearc:
Can I get your opinions on #328?

riftsin:
#328 looks like a better alternative, closing this issue :)

PR #322, opened 2024-03-07
fix(prettier): Buffer range formatting

cryptomilk:
Currently we provide a filename, which means prettier will format the given file and not what we feed via stdin.


stevearc:
Good catch, thanks!

cryptomilk:
Thanks. It still doesn't work as expected as there is a bug in prettier: https://github.com/prettier/prettier/issues/16140

PR #325, opened 2024-03-09
fix: add cwd to honor project php-cs-fixer

zerodahero:
This adds the `cwd` option into the default php-cs-fixer options in order to honor project-level configuration.

`composer.json` should be the "safest" project root option here, since if a project doesn't have a `.php-cs-fixer.dist.php` or similar config file, it looks like we may accidentally use a parent config instead of the default config (currently used with these conform settings), which I believe would be different than typical php-cs-fixer behavior (i.e. use project-level config, or default)


stevearc:
LGTM thanks!

PR #326, opened 2024-03-10
feat: improved swiftformat configuration

wojciech-kulik:
I noticed that `swiftformat` configuration is missing some details. 

I added support for range formatting + I specified how to find cwd.


stevearc:
Thanks for the improvements!

PR #328, opened 2024-03-13
fix(rustfmt): use Cargo.toml settings and default to recent edition

### LINKED ISSUE #320: bug: rust async code formatting error

stevearc:
An alternative to #321
closes #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,6 +504,7 @@ require("conform").formatters.my_formatter = {
 
 - [injected](doc/formatter_options.md#injected)
 - [prettier](doc/formatter_options.md#prettier)
+- [rustfmt](doc/formatter_options.md#rustfmt)
 
 <!-- /FORMATTER_OPTIONS -->
 
--- a/doc/formatter_options.md
+++ b/doc/formatter_options.md
@@ -4,6 +4,7 @@
 
 - [injected](#injected)
 - [prettier](#prettier)
+- [rustfmt](#rustfmt)
 
 <!-- /TOC -->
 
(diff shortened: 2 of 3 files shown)


riftsin:
Oh that's really cool, looks better than #321.
Question, does the ctx passed on lua/conform/formatters/rustfmt.lua:12 refers to the current buffer ? (learning :) )

stevearc:
Yes, the `ctx` is created for the specific buffer being formatted. You can find the definition of that object here:
https://github.com/stevearc/conform.nvim/blob/db2c697fe8302f0328b50b480204be1b577a1e2f/lua/conform/init.lua#L47-L51

PR #338, opened 2024-03-16
Remove --force flag in sqlfluff since it's default now

petobens:
Otherwise we get the following error `The -f/--force option is deprecated as it is now the default behaviour.`


stevearc:
LGTM thanks!

PR #344, opened 2024-03-18
feat: support crystal tool format

hougesen:
Adds support for the official [crystal](https://crystal-lang.org/) formatter.


stevearc:
Thanks for the PR!

PR #350, opened 2024-03-26
feat(prettier, prettierd): add mjs files to supported config files

Nitestack:
Prettier supports ESM in their configuration, so I think it should be added to the supported config files.

**Commit:** https://github.com/prettier/prettier/commit/44d246a2a25473f65ab6be98cf5fe63d3ffe8f8d


stevearc:
LGTM thanks!

PR #351, opened 2024-03-26
Add formatter for Inko

yorickpeterse:
This adds support for formatting Inko (https://inko-lang.org/) source files using the `inko fmt` command.


yorickpeterse:
The `inko fmt` command isn't available yet in the latest stable release of Inko, but merging this ahead of time shouldn't pose a problem I think, because in the worst case conform will log some warning and just ignore the non-existing sub command. I'm also happy with just waiting with merging this PR until the release is published :smiley:

stevearc:
LGTM thanks for the PR!

Issue #360, opened 2024-04-02
when prettier is configed to use crlf eol, unexpected empty line is appended to the file after conform formatting


li6in9muyou:
### Neovim version (nvim -v)

nvim 0.9.5

### Operating system/version

Windows 11

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

# before my fix
```
22:50:06[DEBUG] Running formatters on D:\projects\conform-prettier-crlf-reproduce-bug\index.html: { "prettier" }
22:50:06[INFO] Run prettier on D:\projects\conform-prettier-crlf-reproduce-bug\index.html
22:50:06[TRACE] Input lines: { "<body>", "  <div></div>", "</body>" }
22:50:06[DEBUG] Run command: { "prettier.cmd", "--stdin-filepath", "D:\\projects\\conform-prettier-crlf-reproduce-bug\\index.html" }
22:50:06[DEBUG] Run CWD: D:/projects/conform-prettier-crlf-reproduce-bug
22:50:06[DEBUG] prettier exited with code 0
22:50:06[TRACE] Output lines: { "<body>\r", "  <div></div>\r", "</body>\r" }
22:50:06[TRACE] prettier stderr: { "" }
22:50:06[TRACE] Applying formatting to D:\projects\conform-prettier-crlf-reproduce-bug\index.html
22:50:06[TRACE] Comparing lines { "<body>", "  <div></div>", "</body>" } and { "<body>\r", "  <div></div>\r", "</body>\r" }
22:50:06[TRACE] Diff indices { { 1, 3, 1, 3 } }
22:50:06[TRACE] Applying text edits: { {
    newText = "\r\n  <div></div>\r\n</body>\r",
    range = {
      ["end"] = {
        character = 7,
        line = 2
      },
      start = {
        character = 6,
        line = 0
      }
    }
  } }
22:50:06[TRACE] Done formatting D:\projects\conform-prettier-crlf-reproduce-bug\index.html
```

# after my fix

```
22:50:03[DEBUG] Running formatters on D:\projects\conform-prettier-crlf-reproduce-bug\index.html: { "prettier" }
22:50:03[INFO] Run prettier on D:\projects\conform-prettier-crlf-reproduce-bug\index.html
22:50:03[TRACE] Input lines: { "<body>", "  <div></div>", "</body>" }
[...]

Issue #274, opened 2024-01-13
bug: Blank line at EOF with powershell_es


MarcoBuess:
### Neovim version (nvim -v)

0.9.4

### Operating system/version

Windows 10.0.19045

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

```console
Log file: C:\Users\****\AppData\Local\nvim-data\conform.log
          13:09:11[DEBUG] Running LSP formatter on c:\****\****\Set-Up.ps1
          13:09:11[DEBUG] Converting full-file LSP format to piecewise format
          13:11:29[DEBUG] Running LSP formatter on c:\****\test.ps1
          13:11:29[DEBUG] Converting full-file LSP format to piecewise format
          13:11:31[DEBUG] Running LSP formatter on c:\****\test.ps1
          13:11:31[DEBUG] Converting full-file LSP format to piecewise format
          13:11:47[DEBUG] Running LSP formatter on c:\****\test.ps1
          13:11:47[DEBUG] Converting full-file LSP format to piecewise format

Formatters for this buffer:
LSP: powershell_es

Other formatters:
stylua ready (lua) C:\Users\****\AppData\Local\nvim-data\mason\bin\stylua.CMD
```

### Describe the bug

When using format on save, a new line gets added at EOF. Disabling `conform.nvim` and formatting with `vim.lsp.buf.format()` doesn't show this behavior.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. Open `.ps1` file
2. Save with `format_on_save` enabled.

### Expected Behavior

Code gets formatted but no new line added at EOF.

### Minimal example file

Write-Output -InputObject "Hello, World!"

### Minimal init.lua

```Lua
Minimal lazy spec:

{
    "stevearc/conform.nvim",
    dependencies = { "williamboman/mason.nvim" },
    event = { "BufWritePre" },
    cmd = { "ConformInfo" },
    opts = {
        formatters_by_ft = {
            lua = { "stylua" },
        },
        format_on_save = { timeout_ms = 500, lsp_fallback = true },
        log_level = vim.log.levels.DEBUG,
    },
},
```

MarcoBuess:
[...]

PR #361, opened 2024-04-02
fix: handle windows line ending when config.stdin is true

li6in9muyou:
improve the fix of issue #274 and fix issue #360


stevearc:
LGTM. Thanks for the fix!

PR #373, opened 2024-04-13
fix(biome-check): use safe fixes

cesconix:
This PR updates the **biome-check** formatter. The modification changes the default behavior from using unsafe code fix mode to safe code fix mode to enhance reliability and security when auto-correcting code. 

More details about the tool can be found at [biomejs.dev/linter](https://biomejs.dev/linter/).


stevearc:
Yeah, I think this makes sense. Safe should be the default, and if people want to apply the unsafe fixes they can either customize this formatter or clone it to create a separate one.

PR #378, opened 2024-04-16
Correctly find prettierd executable on windows

dimhatz:
When formatting with `prettierd` on windows, even though the executable exists, there is the following error:

```
19:50:56[ERROR] Formatter 'prettierd' error in jobstart: Vim:E903: Process failed to start: no such file or directory: "C:/Users/myuser/mytest/node_modules/.bin/prettierd"
```

This PR adds detection and selection of `.cmd` executable on windows, the same way we do in `prettier.lua`, which fixes the above.


stevearc:
LGTM thanks!

PR #383, opened 2024-04-19
feat: add dos2unix converter script

tanj:
Add dos2unix formatter.

This is convenient simple formatter that fixes when a formatter (such as isort on windows) changes the line endings of streamed text.


stevearc:
Can you take a look at #395 as an alternative solution? I believe this approach is cleaner than adding a new formatter.

tanj:
yes, your solution in #395 resolves my issue and I agree that is a better solution

PR #395, opened 2024-05-06
fix(isort): explicitly pass line endings

stevearc:
Alternative to #383


tanj:
this resolves my issue better than a new filter

PR #402, opened 2024-05-13
Add a generic hcl formatter (hclfmt)

jonasohland:
Adds a generic formatter for `hcl` files.

I realize that there are already 2 formatters for hcl files defined in this repo, but to use either one of them, you need to have the specific tool installed (terraform or packer).

Meant to be used with the `hclfmt` binary that is built from the [hcl project](https://github.com/mason-org/mason-registry/blob/main/packages/hclfmt/package.yaml). This is also the one that can be obtained with mason: [registry link](https://github.com/mason-org/mason-registry/blob/main/packages/hclfmt/package.yaml).

Formats input from stdin and writes to stdout.

### DIFF
--- a/lua/conform/formatters/hcl.lua
+++ b/lua/conform/formatters/hcl.lua
@@ -2,7 +2,7 @@
 return {
   meta = {
     url = "https://github.com/hashicorp/hcl",
-    description = "A formatter for HCL files",
+    description = "A formatter for HCL files.",
   },
   command = "hclfmt",
 }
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for the PR!

lamchau:
in case someone was looking on how to add this as a standalone and to save some time - since i was expecting

```lua
hcl = { "hclfmt" }
``` 

since some/most formatters match the binary (maybe we'd rename `hcl.lua` to `hclfmt.lua` (but maybe not since there are multiples of that)

```lua
require("conform").setup({
  formatters_by_ft = {
    hcl = { "hcl" },
  }
});
```

install via just `go`
```sh
go install github.com/hashicorp/hcl/v2/cmd/hclfmt@latest
```

PR #405, opened 2024-05-15
feat: add support for bpfmt

yoofa:


yoofa:
bpfmt help command:
$ bpfmt -h

usage: bpfmt [flags] [path ...]
  -d	display diffs instead of rewriting files
  -l	list files whose formatting differs from bpfmt's
  -o	write result to stdout
  -s	sort arrays
  -w	write result to (source) file

use -w to change file directly.
 if use -o, should change stdin = true

stevearc:
Thanks for the PR!

PR #410, opened 2024-05-17
docs: extract some function arguments into their own classes to get `per field` autocompletion + documentation

titaniumtraveler:
I was kinda annoyed by the autocompletion not working there, so I fixed it.

The way I did it, the `opts` documentations are duplicated between the classes
and the opts of the function, which is kinda sad.
The only good solution for this I see would be if lua_ls had support for `inline`
definitions of classes.


stevearc:
This is something that's bothered me for a while: the difficulty of defining class or complex table types inline as function parameters. The current format is completely custom (not handled at all by LuaLS or other tools) and designed to be easy to parse for documentation generation. This finally gave me the push I needed to add some additional capabilities to the documentation generator; it can now reference class definitions and automatically inline them when generating function docs. This lets us remove the duplicate inline class definition on the function.

The only downside of this is that now it's a bit less obvious what the parameters are just from reading the code, since the object is no longer inlined in the comments. I think this is a fair trade for better typechecking.

Thanks for the PR!

Issue #363, opened 2024-04-04
Using `leptosfmt`


Rolv-Apneseth:
Hey, not an issue just thought I would leave this here, both in case you'd like to include it as a recipe or just for other people who come looking for this, but this is the snippet I'm using to get formatting for [leptos](https://github.com/leptos-rs/leptos) projects to work using [leptosfmt](https://github.com/bram209/leptosfmt):

```lua
formatters = {
    rustfmt = {
        command = "leptosfmt",
        args = {
            "--stdin",
            "--rustfmt",
        },
    },
}
```

This replaces `rustfmt` with `leptosfmt`, which will still call `rustfmt` due to that `--rustfmt` flag so it still formats code outside of `leptos` projects just fine too.

PR #415, opened 2024-05-22
feat: add support for leptosfmt

clispios:
As title says: this adds support for leptosfmt 😀 

This is not the same as #363 since that suggestion does both leptosfmt and rustfmt combined. This commit just adds a leptos formatter, which can then be chained together sequentially with rustfmt in a conform config, if someone wants to do that.


stevearc:
Thanks for the PR!

PR #418, opened 2024-05-22
feat: add `ruff_organize_imports` formatter

mehalter:
This adds another `ruff` based formatter which tells it to organize imports. Useful for replacing `isort` with ruff. Neither the `ruff_fix` nor `ruff_format` cover this case or include it in their formatting.

Thanks in advance!


stevearc:
LGTM thanks!

Issue #252, opened 2023-12-26
Formatting with prettier with Windows path with spaces not working


GitMurf:
If I have a space in my path at all of my git repository (where prettier is installed in node_modules) then I get the following error:

![image](https://github.com/stevearc/conform.nvim/assets/64155612/4caf78f5-6705-402e-9a03-4733245efedf)

Notice the actual start to my path to my git repo locally has a space here which is exactly where the error above cutoff.

![image](https://github.com/stevearc/conform.nvim/assets/64155612/a2ed4b4a-bbd8-490b-88ad-8b0adc3b889a)

_Originally posted by @GitMurf in https://github.com/stevearc/conform.nvim/issues/236#issuecomment-1868100641_

GitMurf:
Above was from a couple days ago. Below is still from today after the recent update (still not working):

![image](https://github.com/stevearc/conform.nvim/assets/64155612/b35d29f9-2a33-47a1-bb91-2c21696ed140)

![image](https://github.com/stevearc/conform.nvim/assets/64155612/a4c320ae-312d-4dde-af9d-28760e76a91b)

GitMurf:
Thanks for all your hard work on this plugin :) let me know if you need anymore details.

GitMurf:
curious if you have had a chance to look at this yet? also I am happy to take a look at doing a PR to try and fix it if you are able to point me in the right direction. I am a TS developer and brand new to neovim / lua etc. so I am capable but just need a little direction of where to look and then I could play around and do some testing to see if I can fix it for Windows. I tried on my own the other day for a couple hours but kept running in circles and couldn't quite figure out where everything flowed through to properly fix it.

stevearc:
I am unfortunately very time limited these days, and I have a lot of projects to manage, so I have to aggressively triage and carefully choose what I spend time on. I try to make sure I always review PRs, and anything breaking enough to be a P0, but everything else is no guarantees.

Some tips for investigating this yourself: get some detailed log information
```lua
require("conform").setup({
  log_level = vim.log.levels.TRACE
})
```

Try to get an invocation of the formatter on the command line that _does_ work, then see what the difference is between that and the command that conform is using. Gradually mutate one into the other until you can pinpoint what the issue is with the command.

If the commands are the same, then possibly there's some problem with the details of how Neovim is running the command. See if you can replicate the issue with a minimal lua script that calls `vim.fn.jobstart` directly.

msabathier:
[...]

PR #421, opened 2024-05-25
Account for spaces in the file path when formatting with stylua

msabathier:
Hi,

I found a bug when formatting a file that has spaces in its path using stylua on Windows.
The command that was being executed was something like:
`stylua --search-parent-directories --stdin-filepath C:\\some\\path with\\spaces.lua`
This was causing stylua to search for `C:\\some\\path` and `with\\spaces.lua` as two separate files and crash.
I simply added simple quotes around the $FILENAME environment variable so the executed command would be:
`stylua --search-parent-directories --stdin-filepath 'C:\\some\\path with\\spaces.lua'`

I don't think this should break anything but i didn't do much testing.
Let me know if you want me to look at some particular test cases.

### DIFF
--- a/lua/conform/formatters/stylua.lua
+++ b/lua/conform/formatters/stylua.lua
@@ -6,13 +6,13 @@ return {
     description = "An opinionated code formatter for Lua.",
   },
   command = "stylua",
-  args = { "--search-parent-directories", "--stdin-filepath", "$FILENAME", "-" },
+  args = { "--search-parent-directories", "--stdin-filepath", "'$FILENAME'", "-" },
   range_args = function(self, ctx)
     local start_offset, end_offset = util.get_offsets_from_range(ctx.buf, ctx.range)
     return {
       "--search-parent-directories",
       "--stdin-filepath",
-      "$FILENAME",
+      "'$FILENAME'",
       "--range-start",
       tostring(start_offset),
       "--range-end",


stevearc:
I find this very surprising. From `:help jobstart`

> jobstart({cmd} [, {opts}])
> If {cmd} is a List it runs directly (no 'shell').

That means that quoting shouldn't matter at all because it's not getting parsed by a shell. What happens if you run this directly with `jobstart`?

```lua
local jid = vim.fn.jobstart(
  { "stylua", "--search-parent-directories", "--stdin-filepath", "C:\\some\\path with\\spaces.lua" },
  {
    stdout_buffered = true,
    stderr_buffered = true,
    stdin = "pipe",
    on_stdout = function(_, data)
      print("STDOUT:", table.concat(data, "\n"))
    end,
    on_stderr = function(_, data)
      print("STDERR:", table.concat(data, "\n"))
    end,
    on_exit = function(_, code)
      print("EXIT:", code)
    end,
  }
)

local buffer_text = [[
local foo  = 1
]]
vim.api.nvim_chan_send(jid, buffer_text)
vim.fn.chanclose(jid, "stdin")
```

LifeAdventurer:
the whitespace in the account name is a big issue... cannot use any of the formatters...

msabathier:
Okay so this goes deeper than i thought, it seems to be an issue with Mason because it works fine if i expose the `stylua.exe` to conform rather than the Mason `stylua.cmd`.
I will be looking into it and report back my findings in issue #252.
In anycase, my PR is not the way to go so i will close it.

PR #422, opened 2024-05-25
fix(dprint): add support node_modules

akineko:
[Install - dprint - Code Formatter](https://dprint.dev/install/)
[dprint - npm](https://www.npmjs.com/package/dprint)

dprint is supports installation via npm, so I have also included node_modules in the search targets.


stevearc:
LGTM thanks!

PR #425, opened 2024-05-29
[feat]: add treesitter query formatter

magnuslarsen:
I found out that nvim-treesitter comes with a formatter; that they themselves use to format their queries:\
https://github.com/nvim-treesitter/nvim-treesitter/blob/ea2b137f35fb1e87a6471ec311805920fdf45745/.github/workflows/lint.yml#L36-L51

They maintain it, and it [ships with the plugin](https://github.com/nvim-treesitter/nvim-treesitter/blob/ea2b137f35fb1e87a6471ec311805920fdf45745/scripts/format-queries.lua)... So why not use it?

I threw this a little hastily together; if you have some better ideas for the condition or anything else, please let me know :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,8 +221,8 @@ You can view this list in vim with `:help conform-formatters`
 - [fish_indent](https://fishshell.com/docs/current/cmds/fish_indent.html) - Indent or otherwise prettify a piece of fish code.
 - [fixjson](https://github.com/rhysd/fixjson) - JSON Fixer for Humans using (relaxed) JSON5.
 - [fnlfmt](https://git.sr.ht/~technomancy/fnlfmt) - A formatter for Fennel code.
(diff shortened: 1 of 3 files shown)


stevearc:
This is fantastic, thanks for the PR!

ribru17:
Thank you! For anyone using the latest `nvim-treesitter`, the `query` parser used to format the files got some breaking changes, so I had to use the following to get it to work (recognize the latest parser rather than the one bundled with neovim):
```lua
local qfmt = require('conform.formatters.format-queries')
local nvim_ts_path = vim.fn.stdpath('data') .. '/lazy/nvim-treesitter/'
util.add_formatter_args(qfmt, { '-c', 'set rtp^=' .. nvim_ts_path })
```
NOTE: This will only work when using lazy but you can replace `nvim_ts_path` with your path to nvim-treesitter and it will work

stevearc:
@ribru17 thanks for the tip! I've incorporated that into the formatter

ribru17:
Awesome, thank you!

magnuslarsen:
Thanks both of you!

PR #434, opened 2024-05-29
feat(rst): support rstfmt

hougesen:
Not to be confused with rustfmt

### DIFF
--- a/lua/conform/formatters/rstfmt.lua
+++ b/lua/conform/formatters/rstfmt.lua
@@ -1,0 +1,10 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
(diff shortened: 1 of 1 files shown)


stevearc (changes requested):
I'm confused because the command used here is `rstfmt` but the linked repo is for `docstrfmt` and calls out `rstfmt` as a separate program. What is the intention here?

hougesen:
Ah shit yeah, my bad 😅

I was just running through all the formatters that [mdsf](https://github.com/hougesen/mdsf) has, that conform was missing.

`docstrfmt` being first on google threw me off

image

stevearc:
Checks out, thanks!

PR #440, opened 2024-06-02
unsaved files: smarter file name extension

emmanueltouzery:
the extension used to be always the file type, be smarter now. For instance prettier wouldn't recognize a `.typescript` extension.

### DIFF
--- a/lua/conform/ft_to_ext.lua
+++ b/lua/conform/ft_to_ext.lua
@@ -1,0 +1,13 @@
+return {
+  elixir = "ex",
+  graphql = "gql",
+  javascript = "js",
+  javascriptreact = "jsx",
+  markdown = "md",
+  perl = "pl",
+  python = "py",
+  ruby = "rb",
+  rust = "rs",
+  typescript = "ts",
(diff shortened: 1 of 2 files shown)


emmanueltouzery:
to be clear, a reproduction would be..

create an empty unsaved file...
type `function() {return 5;}`
set ft=typescript
require('conform').format() 

before the change, nothing would get indented, as prettier says it doesn't recognize the .typescript extension, after the change it works.

stevearc:
Sure, the approach looks good! The one request would be to pull the map out of the function. I think it makes sense to just put it in its own file, something like `ft_to_ext.lua`.

emmanueltouzery:
I updated the PR. thinking about it some more, i actually think for many file formats, the extension is not very important.. it's probably mostly important for prettier for instance, which knows how to format many file types.

but presumably it can't hurt to get the file extension right. for many file types there is nothing to do (for instance lua, sql, html, zig..). I added those that I noticed.

stevearc:
LGTM thanks for the PR!

PR #445, opened 2024-06-06
feat: add support for d2

rtgiskard:
As `d2` filetype is not defined for nvim at the time, you may need to `set filetype=d2` for this to work


stevearc:
Thanks for the PR!

PR #446, opened 2024-06-07
feat(latex): support llf

engeir:
Adds support for the [llf](https://repo.or.cz/llf.git) Latex reformatter / beautifier.


stevearc:
Thanks for the PR!

PR #451, opened 2024-06-09
feat: add support for for vsg

fredrikfoss:
Hi, love your project. This adds support for the [VHDL Style Guide](https://github.com/jeremiah-c-leary/vhdl-style-guide) (VSG) formatter.

The formatter doesn't auto-detect configuration files, so I added functionality to look for `.vsg.yaml` and `.vsg.json` in and under the file's directory and in `xdg-config/vsg` and `~/.config/vsg`. I couldn't find any mention of a standard name for said files in the documentation, but I saw some other people using these names and is also what I use myself. This might be out of scope, but it is quite nice to have, at least until they settle on something themselves or bake in the functionality. But please let me know what you think.

### DIFF
--- a/lua/conform/formatters/vsg.lua
+++ b/lua/conform/formatters/vsg.lua
@@ -1,19 +1,33 @@
-local function find_local_config()
-  local current_file = vim.api.nvim_buf_get_name(0)
-  local local_configs = { ".vsg.yaml", ".vsg.yml", ".vsg.json" }
-  return vim.fs.find(local_configs, {
-    path = vim.fs.dirname(current_file),
-    upward = true,
-  })[1]
-end
+local config_files = {
+  "vsg_config.yaml",
+  "vsg_config.yml",
+  "vsg_config.json",
+  "vsg.yaml",
+  "vsg.yml",
+  "vsg.json",
+  ".vsg_config.yaml",
+  ".vsg_config.yml",
+  ".vsg_config.json",
+  ".vsg.yaml",
+  ".vsg.yml",
+  ".vsg.json",
+}
 
-local function find_global_config()
-  local xdg_config_home = os.getenv("XDG_CONFIG_HOME") or os.getenv("HOME") .. "/.config"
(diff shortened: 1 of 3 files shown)


lua/conform/formatters/vsg.lua
```diff
@@ -0,0 +1,38 @@
+local function find_local_config()
+  local current_file = vim.api.nvim_buf_get_name(0)
+  local local_configs = { ".vsg.yaml", ".vsg.yml", ".vsg.json" }
+  return vim.fs.find(local_configs, {
+    path = vim.fs.dirname(current_file),
```
stevearc, line 5: Instead of using the current buffer, the args function should take the parameters `(self, ctx)` and you should use `ctx.dirname` for this

lua/conform/formatters/vsg.lua
```diff
@@ -0,0 +1,38 @@
+local function find_local_config()
+  local current_file = vim.api.nvim_buf_get_name(0)
+  local local_configs = { ".vsg.yaml", ".vsg.yml", ".vsg.json" }
```
stevearc, line 3: Since the official docs use `vsg_config.yaml` [in their examples](https://vhdl-style-guide.readthedocs.io/en/latest/editor_integration/vim.html#neovim), it probably makes sense to add support for that + variations as well.

stevearc (changes requested):
Kind of baffling that a formatter supports config files but doesn't automatically look for one with a standard name. The approach looks like a reasonable workaround; left a few recommendations for small improvements

fredrikfoss:
Yeah, it's unfortunate. It even supports a ton of configuration. But it's more or less the only vhdl formatter available aside from buggy web-based ones, so can't complain too much. Thanks for the pointers; it now uses `ctx.dirname`. I also made a few more changes when i first started, please have a look. It now supports `[.]vsg_config.{yaml,yml,json}` and `[.]vsg.{yaml,yml,json}`, both locally and in xdg_config. That's quite a few options. Let me know what you think.

stevearc:
LGTM thanks for the PR!

PR #461, opened 2024-06-17
doc: add LSP annotations for `lsp_format` options

mehalter:
This utilizes the Lua language server annotations to document each option and improve the code completion.

Documentation of the option overall:
2024-06-17_11:24:11_screenshot

Completing an option:
2024-06-17_11:24:25_screenshot


mehalter:
Also added a commit to fix the type annotation that a function for `format_on_save` and `format_after_save` can return `nil` to not format. I guess technically it could also be "false", but seems more like a lua interpreter side effect and not really the intended use.

mehalter:
Ok! No problem! Do you want me to open a different PR with the other annotation fix in the meantime? Or you could commit it separately or it can wait until your doc tooling support.

stevearc:
Yeah, I'll make a quick commit to fix the annotations, and will look into the type alias parsing when I get a chance

stevearc:
docs support added. Thanks for the PR!

PR #462, opened 2024-06-19
fix: correctly call `lsp_format = 'fallback'`

dpetka2001:
This fixes the `fallback` behavior to use it when no other formatter has been used, instead of checking for filetype formatters as well. I tested it locally and it seems to work without any side-effects. 

Please feel free to disregard this if you think a better approach is available or this PR introduces some side-effects that I can't think of.

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -451,6 +451,9 @@ M.format = function(opts, callback)
   elseif not vim.tbl_isempty(formatters) then
     run_cli_formatters(handle_result)
     return true
+  elseif opts.lsp_format == "fallback" and M.will_fallback_lsp(opts) then
+    lsp_format.format(opts, callback)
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for the fix! This does fix the issue, but I went with bde3bee1773c96212b6c49f009e05174f932c23a instead. I think it is a "more correct" way to address the problem

dpetka2001:
I guess [M.will_fallback_lsp](https://github.com/stevearc/conform.nvim/blob/bde3bee1773c96212b6c49f009e05174f932c23a/lua/conform/init.lua#L701-L709) is not used any more?

stevearc:
`M.will_fallback_lsp` is not used anymore. It is preserved to retain backwards compatibility, but we don't call it directly.

PR #464, opened 2024-06-19
fix: ignore empty opts.formatters like nil

Frederick888:
Same rationale as [1]. When there are no LSP formatters and opts.formatters = {}, we should normally call
list_formatters_for_buffer.

[1] https://github.com/stevearc/conform.nvim/commit/bde3bee1773c96212b6c49f009e05174f932c23a

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -378,6 +378,9 @@ M.format = function(opts, callback)
   local lsp_format = require("conform.lsp_format")
   local runner = require("conform.runner")
 
+  if is_empty_table(opts.formatters) then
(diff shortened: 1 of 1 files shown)


Frederick888:
I guess I somehow managed to confuse myself in #463 lol...

stevearc:
Hmmm...I'm not sure about this one. The normal behavior is that if you explicitly pass in `formatters`, we ignore the `formatters_by_ft` and only use the formatters you passed in. This would change that behavior so that we do use the `formatters_by_ft` if `formatters` is passed in _but_ is an empty table. What is the reasoning behind this change?

PR #467, opened 2024-06-23
feat(djlint): respect local buffer tabstop config

### LINKED ISSUE #457: feature request(djlint): respect editorconfig/local buffer `tabstop`

### Describe the feature

`djlint` could format buffer with [indent](https://djlint.com/docs/configuration/#indent) config based on editorconfig/local buffer `tabstop`

### Provide background

I found that `djlint` supports [indent](https://djlint.com/docs/configuration/#indent) config argument. It would be nice if this could be incorporated to `conform` config so that it could run `djlint` with config based on local buffer `tabstop` which is set based on editorconfig.

nice to have

### Additional details

I found a bit similar feature in `markdown-toc` [config](https://github.com/stevearc/conform.nvim/blob/797d1f622a23d4a21bb58218bdf5999a9beac4ef/lua/conform/formatters/markdown-toc.lua#L12) that could respect local buffer `tabstop` (although it's just check whether it's space or tab).

I would like to try to implement this feature so any guidance/suggestion is welcome.

Additionally, I think this feature could be implemented to all formatters that support indent config like this (however I have no idea what the formatters are).

mangkoran:
Formatter now able to use local buffer tabstop config. One of the case is when the tabstop is set by editorconfig.

Close #457

### DIFF
--- a/lua/conform/formatters/djlint.lua
+++ b/lua/conform/formatters/djlint.lua
@@ -6,10 +6,10 @@ return {
     description = "✨ HTML Template Linter and Formatter. Django - Jinja - Nunjucks - Handlebars - GoLang.",
   },
   command = "djlint",
-  args = {
-    "--reformat",
-    "-",
-  },
+  args = function(_, ctx)
+    local indent = vim.bo[ctx.buf].tabstop or 4 -- default is 4
+    return { "--reformat", "--indent", indent, "-" }
+  end,
   cwd = util.root_file({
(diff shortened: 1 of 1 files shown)


mangkoran:
Additionally I would like to add a section in the docs to mention that user could config their formatter to respect local buffer ts like this if the formatter support indentation config through argument.

stevearc:
LGTM thanks!

PR #481, opened 2024-06-28
feat(shfmt): add automatic indentation detection

mehalter:
This adds the automatic setting of `-i` so that `shfmt` follows the current indentation settings of neovim. Whether it is manually set by the user or set through external factors such as `.editorconfig`

I also checked to see if this is done elsewhere for other formatters and found a couple others which I was able to improve the "correctness" of their indentation size calculations

EDIT: I was thinking about this and it seems like having the effective `shiftwidth` value is quite nice. And because this seems to be "contextual information" it seemed natural to put it in the `conform.Context` type since it is virtually free to calculate performance wise. Let me know if you disagree. Another approach would be to make some sort of utility module that is imported by formatters on an as needed basis that calculates this. This is basically the same as `vim.fn.shiftwidth` (`:h shiftwidth()`) with the only difference being that it calculates it specifically for the buffer in the context rather than the "current buffer". This allowed for a very nice refactor across the formatters that utilize this field and is available for future formatters. Also nice to make sure we have a common method for calculating indentation size.

Also `shiftwidth` and `tabstop` are always integers according to the neovim docs so no need to have default fallbacks or check for `nil`

### DIFF
--- a/lua/conform/formatters/djlint.lua
+++ b/lua/conform/formatters/djlint.lua
@@ -7,8 +7,12 @@ return {
   },
   command = "djlint",
   args = function(_, ctx)
-    local indent = vim.bo[ctx.buf].tabstop or 4 -- default is 4
-    return { "--reformat", "--indent", indent, "-" }
+    local bo = vim.bo[ctx.buf]
+    local indent_size = bo.shiftwidth
+    if indent_size == 0 or not indent_size then
+      indent_size = bo.tabstop or 4 -- default is 4
+    end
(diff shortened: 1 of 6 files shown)


mehalter:
If this is a common task, it might be useful to provide this maybe as a utility so formatters can easily calculate it on demand or we could even put this as part of the formatter context that is passed to the `args` and `prepend_args` functions

stevearc:
Nice QOL improvement for people writing formatters. Thanks for the well reasoned PR!

PR #482, opened 2024-06-29
fix(biome-check): use --write instead of deprecated --apply

abellaismail7:
biome deprecated --apply.
```

--apply   Alias for `--write`, writes safe fixes, formatting and import sorting 
          (deprecated, use `--write`)

```


stevearc:
LGTM thanks!

PR #484, opened 2024-07-01
Adds Swiftlint formatter

jakenvac:
I've added [swiftlint](https://github.com/realm/SwiftLint) as a formatter, I've tested over the weekend and it works great.

The command used to format is as follows:
```sh
swiftlint lint \
  --use-stdin \ # take input from stdin
  --fix \ # automatically fix errors where possible
  --format # apply sourcekit lsp default formatting if available
```


stevearc:
Thanks for the PR!

PR #488, opened 2024-07-07
feat: add new format option: undojoin

Grolaf:
This option allow user to automatically perform `undojoin` command before formatting. This is useful if the user uses an autosave plugin + format on save, because in case of undo, it will undo the last change AND the formatting. Without this option, it will only undo the formatting.


stevearc:
This is a nice feature addition. Thanks for the PR!

Grolaf:
@vuthanhtung2412 could you please share your config ? 
Actually, mine is defined like this : 

```
return {
  "stevearc/conform.nvim",
  optional = true,
  opts = {
    format = {
      undojoin = true,
    },
    formatters_by_ft = {
      python = { "black" },
      typescript = { "prettierd" },
      typescriptreact = { "prettierd" },
      json = { "prettierd" },
    },
  },
}
```

And it's working like a charm.

BTW, sorry for answering this so lately.

PR #491, opened 2024-07-15
feat!: large rework of configuration logic

stevearc:
This PR bundles several changes related to how conform can be configured. There are two breaking changes in the form of deprecating a method and deprecating the formatter alternation syntax, however both will continue to work for now after one warning notification. The final breaking change is dropping support for Neovim 0.8, which was done because the tests now fail on 0.8 due to nvim-treesitter. Since 0.9 has been around for a while, it's easier to just bump the requirement and bundle that breaking change with these others.

### Default format options
**feature**: New `default_format_opts` option that will allow you to configure the default behavior of `conform.format()`
**motivation**: The primary reason for this is because the per-filetype configuration (below) needs this to make sense. If we can configure the `lsp_format` on a per-filetype basis, we also probably want to be able to customize the _default_ value to avoid repetition.
**example**:
```lua
require("conform").setup({
  default_format_opts = {
    lsp_format = "fallback",
    timeout_ms = 500,
  },
})
```

### Per-filetype configuration of format options
**feature**: `formatters_by_ft` can now customize some format options
**motivation**: make it easier to customize behavior for specific filetypes. Previous advice for changing the LSP fallback behavior for one filetype was fairly involved. People frequently perform formatting in two places: on save, and on a particular keymap (e.g. `<leader>f`). It can be annoying to keep the behavior in those two places in sync. This, combined with `default_format_opts`, should make it much easier to keep the logic simple in both places.
**example**:
```lua
require("conform").setup({
  formatters_by_ft = {
    go = { "goimports", "gofmt", lsp_format = "never" },
    rust = { "rustfmt", lsp_format = "fallback" },
    javascript = { "prettierd", "prettier", timeout_ms = 2000, stop_after_first = true },
  },
  format_on_save = {},
})
vim.keymap.set("", "<leader>f", function()
  require("conform").format({ async = true })
end, { desc = "Format code" })
```

### Notification for no available formatters
**feature**: You will now get a once-per-filetype notification when you call `conform.format` on a buffer that _does_ have formatters configured when none of those formatters are available (i.e. they aren't installed). There is an option to disable this notification.
**motivation**: #480
**example**:
```lua
require("conform").setup({
  -- to disable these new notifications
  notify_no_formatters = false,
})
```

### Deprecate formatter alternation syntax for option to only run the first available formatter
**feature**: Previously you could use nested list syntax like `{ { "prettierd", "prettier" } }` to run the first formatter available instead of all formatters. Now, `conform.format` takes a single boolean option `stop_after_first` that will determine whether to use _all_ formatters, or just the first available.
**motivation**: This is a useful feature, but the syntax is a weird special case and complicates the data structures we accept and deal with internally. It's also harder for new users to understand what magic is happening.
**example**:
```lua
require("conform").setup({
  formatters_by_ft = {
    -- deprecated syntax
    -- javascript = { { "prettierd", "prettier" } },
    javascript = { "prettierd", "prettier", stop_after_first = true },
  },
})
```

Note that we are aware that this new format is slightly less flexible, since you can't use constructs like `{ { "prettierd", "prettier" }, "injected" }` anymore. This behavior is still possible, you will just need to write it yourself using `get_formatter_info` and checking what is available. We think that the simplification of the code and the common case is worth the added complication for these advanced uses.

### Deprecate `will_fallback_lsp` for `list_formatters_to_run`
**feature**: new API method that will return the exact list of formatters that will be run, accounting for all `stop_after_first`, LSP fallback, and other logic.
**motivation**: the original purpose for `will_fallback_lsp` was for use in statusline plugins to help them determine if the formatters will be used or if conform will use LSP instead. With the refactor of `lsp_format` to add more ways LSP formatting can work, it makes more sense to just bundle all this logic up into a function that just gives you what you want. This new function should make it easier to integrate conform into statuslines.
**example**:
```lua
_G.conform_status = function()
  local formatters, use_lsp = require("conform").list_formatters_to_run()
  local names = vim.tbl_map(function(formatter)
    return formatter.name
  end, formatters)
  if use_lsp then
    table.insert(names, "LSP")
  end
  return "[" .. table.concat(names, ", ") .. "]"
end

vim.o.statusline = [[%!v:lua.conform_status()]]
```


lukoshkin:
```lua
    json = { { "prettierd", "prettier" }, "fixjson" },
    markdown = { { "prettierd", "prettier" }, "markdownlint" },
```
How to implement these settings with the new logic? That is, I want to use fixjson and markdownlint always and make it only optional and first available in the pair "prettierd", "prettier"

stevearc:
@lukoshkin there is a recipe here https://github.com/stevearc/conform.nvim/blob/master/doc/recipes.md#run-the-first-available-formatter-followed-by-more-formatters

PR #500, opened 2024-07-25
Add support for cljfmt

zolotyh:
I added config for yet one Clojure formatting tool. Unfortunately, I haven't found contribution docs, so I tried to use common sense.


stevearc:
Thanks for the PR!

PR #505, opened 2024-07-30
feat(formatters): add nixfmt-rfc-style

ugur-a:
This is different from `nixfmt`, which is currently defined at:
https://github.com/stevearc/conform.nvim/blob/25d48271e3d4404ba017cb92a37d3a681c1ad149/lua/conform/formatters/nixfmt.lua

This new formatter is meant to eventually replace `nixfmt`. More information can be found on [their repo](https://github.com/NixOS/nixfmt)


ugur-a:
On another thought, the same command is used by both versions, so maybe just:
- update the repo link - the current one (https://github.com/serokell/nixfmt) already redirects to the new one (https://github.com/NixOS/nixfmt) 
- update the description - it seems to come from the old repo, which, as I said, doesn't seem to be there anymore

This way, `conform.nvim` will just use the version that's currenty installed

stevearc (changes requested):
I agree, updating the repo + description of the existing formatter sounds like the correct move here

ugur-a:
Done!

stevearc:
Thanks for the PR!

PR #510, opened 2024-08-01
Add support to extensions to arguments.

Bubbet:
I needed this as a workaround to paths that include () on windows for prettierd. Svelte layout groups. A simple enough solution to that problem.

### DIFF
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -48,7 +48,7 @@ M.build_cmd = function(formatter_name, ctx, config)
       :gsub("$FILENAME", ctx.filename)
       :gsub("$DIRNAME", ctx.dirname)
       :gsub("$RELATIVE_FILEPATH", compute_relative_filepath)
-      :gsub("$EXTENSION", ({ ctx.filename:match("(.*)(%..*)") })[2] or "")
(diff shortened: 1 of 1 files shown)


lua/conform/runner.lua
```diff
@@ -48,6 +48,7 @@ M.build_cmd = function(formatter_name, ctx, config)
       :gsub("$FILENAME", ctx.filename)
       :gsub("$DIRNAME", ctx.dirname)
       :gsub("$RELATIVE_FILEPATH", compute_relative_filepath)
+      :gsub("$EXTENSION", ({ ctx.filename:match("(.*)(%..*)") })[2] or "")
```
stevearc, line 51: I think `ctx.filename:match(".*(%..*)$")[1]` would be simpler

stevearc (changes requested):
I'm fine with adding the new variable, but prefer a slight tweak to the match logic

lua/conform/runner.lua
Bubbet: Wasn't sure if you were allowed to do that it's been awhile since I've wrote Lua. Thought the returned values were unpaired already.

stevearc:
Thanks for the PR!

PR #511, opened 2024-08-02
feat: add support for markdownfmt

sieuwerts:
Adding support for markdownfmt. I opted to not include `find_executable` in the command since it's installed via `go install`, and gopath might differ between machines.

I have tested it and it seem to be working fine as long as the executable is reachable from PATH.


stevearc:
Thanks for the PR!

PR #516, opened 2024-08-08
Update README to use sublist for running first available formatter

nebunebu:
```gitdiff
-    -- Conform will run the first available formatter
-    javascript = { "prettierd", "prettier", stop_after_first = true },
+    -- Use a sub-list to run only the first available formatter
+    javascript = { { "prettierd", "prettier" } },
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,8 +131,8 @@ require("conform").setup({
     python = { "isort", "black" },
     -- You can customize some of the format options for the filetype (:help conform.format)
     rust = { "rustfmt", lsp_format = "fallback" },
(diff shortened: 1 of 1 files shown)


stevearc:
This syntax was explicitly deprecated by #491. Why are you trying to bring it back?

nebunebu:
Confusion on my part. I was using an older rev, 25d48271e3d4404ba017cb92a37d3a681c1ad149, and saw that the description for stop after first functionality did not match up on the README.md and `:help conform-options`. My mistake.

PR #517, opened 2024-08-08
Remove indent command line argument for djlint

zidhuss:
When we pass cli arguments to `djlint`, they override those set in the configuration files: https://www.djlint.com/docs/configuration

I'd expect the formatter to respect the configuration file. If a user wants to override they can do so https://github.com/stevearc/conform.nvim/issues/457#issuecomment-2169139931.

This PR reverts the changes made in #467.


stevearc:
Agree that we shouldn't pass arguments by default that can override configuration file settings.

cc @mangkoran

PR #518, opened 2024-08-10
fix(docformatter): update exit_codes, 3 is correct in in-place formatting

ZeinBarhoum:
Fixes error `Formatter 'docformatter' error: unknown error` 

According to the [documentation](https://docformatter.readthedocs.io/en/latest/usage.html) on possible exit_codes: 

> 3 - if any file needs to be formatted (in --check or --in-place mode)

Which is expected when the document needs actual formatting


stevearc:
Thanks for the PR!

Issue #191, opened 2023-11-10
perf: use `uv.spawn` instead of `jobstart` for better child process IO performance


linrongbin16:
Hi, I noticed that the internal implementation of running a formatter command is using the API `jobstart`. It's also the previous implementation of my plugin [linrongbin16/gitlinker.nvim](https://github.com/linrongbin16/gitlinker.nvim), forked from [ruifm/gitlinker.nvim](https://github.com/ruifm/gitlinker.nvim).

I recently added a modified version of `vim.system` (API from latest Neovim nightly), which is a user friendly version of `uv.spawn` (e.g. `vim.loop.spawn` introduced in Neovim 0.5+), it has a super good performance on child process IO. 

> Even when we run a command line and wait for its response (e.g. sync), `uv.spawn` still has the super good performance, compared with `jobstart`.

I'm so lazy so I cannot provide any benchmark data for you, but I feels like my gitlinker.nvim speed up from about 0.8 ~ 1.2 seconds to less than 0.2 seconds. Inside it will run about 2 ~ 5 git commands in a git repo.

If you have any interest, please see:

1. replace old `jobstart` API to `spawn` here: https://github.com/linrongbin16/gitlinker.nvim/commit/35aebb7f4f8d30b7863742864a93cbe0224e8975#diff-78b3d3e35fb07653940e1e29882d42b81f9e83663772c3786617e682cccf29a7L39
2. implement the `spawn` here: https://github.com/linrongbin16/gitlinker.nvim/commit/35aebb7f4f8d30b7863742864a93cbe0224e8975#diff-5b092e9c8aa7a8be4cc1086ddd1101289567012f06c35faa670fbe2578472ae2R229

I could try to submit a PR to improve the plugin performance.

stevearc:
I considered doing this a while ago, but looking at the amount of code that would be required to replace `jobstart` with `spawn`...it's not great. Adding that much code _is_ a noticeable maintenance burden, and I don't want to do it without good reason. Given that I haven't had any complaints about performance, and haven't noticed any performance problems myself (other than the cases where the formatter itself is slow), I don't think there's much upside.

Plus, I get the feeling that nvim 0.10 is fairly close to release. I would expect to see it come out some time in the next couple months. I'd prefer to just wait for that and use `vim.system` then.

lopi-py:
[...]

PR #521, opened 2024-08-14
refactor: use vim.system

lopi-py:
Use `vim.system` instead of `vim.fn.jobstart` for performance gains (#191)

`black` before:
black-before
`black` after:
black-after

### DIFF
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -53,8 +53,8 @@ M.build_cmd = function(formatter_name, ctx, config)
       :gsub("$RELATIVE_FILEPATH", compute_relative_filepath)
       :gsub("$EXTENSION", ctx.filename:match(".*(%..*)$") or "")
     return {
-      vim.o.shell,
-      vim.o.shellcmdflag,
+      vim.split(vim.o.shell, " "),
+      vim.split(vim.o.shellcmdflag, " "),
       string.format("{%s %s}", command, interpolated),
     }
   else
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -359,91 +359,74 @@ local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, c
   if env then
     log.debug("Run ENV: %s", env)
   end
-  local stdout
-  local stderr
   local exit_codes = config.exit_codes or { 0 }
   local jid
-  local ok, jid_or_err = pcall(vim.fn.jobstart, cmd, {
+  local ok, job_or_err = pcall(vim.system, cmd, {
     cwd = cwd,
     env = env,
-    stdout_buffered = true,
-    stderr_buffered = true,
-    stdin = config.stdin and "pipe" or "null",
-    on_stdout = function(_, data)
-      if config.stdin then
-        stdout = data
+    stdin = config.stdin and buffer_text,
+  }, function(result)
+    local code = result.code
+    local stdout = result.stdout and vim.split(result.stdout, "\r?\n") or {}
+    local stderr = result.stderr and vim.split(result.stderr, "\r?\n") or {}
+    if vim.tbl_contains(exit_codes, code) then
+      local output = stdout
+      if not config.stdin then
+        local fd = assert(uv.fs_open(ctx.filename, "r", 448)) -- 0700
+        local stat = assert(uv.fs_fstat(fd))
+        local content = assert(uv.fs_read(fd, stat.size))
+        uv.fs_close(fd)
+        output = vim.split(content, "\r?\n")
       end
-    end,
-    on_stderr = function(_, data)
-      stderr = data
-    end,
-    on_exit = function(_, code)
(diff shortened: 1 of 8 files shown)


lopi-py:
Tests seems to pass for me using Neovim `10.0.1`.
Also I think we need to replace `vim.fn.jobclose`, but I haven't figured it out yet

lopi-py:
It should be ready. I'm using `vim.fn.exepath` for the command, installing a formatter on Windows (😭) which is a batch script (for example with `mason.nvim` or `node`), `uv.spawn` wouldn't recognize it as executable, this is the [same strategy](https://github.com/neovim/nvim-lspconfig/blob/ad32182cc4a03c8826a64e9ced68046c575fdb7d/lua/lspconfig/configs.lua#L22) used by `nvim-lspconfig`

lua/conform/runner.lua
```diff
@@ -359,40 +363,28 @@ local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, c
...
-        stdout = data
-      end
-    end,
-    on_stderr = function(_, data)
-      stderr = data
-    end,
-    on_exit = function(_, code)
+  local ok, job_or_err = pcall(
+    vim.system,
+    cmd,
```
stevearc, line 370: One of the less-documented (possibly undocumented?) features of conform is that if `args` is a string or function that returns a string (as opposed to the normal `string[]`), the `build_cmd` function above will _also_ return a string. Passing a string to `jobstart` will run the command using the shell instead of executing it directly, allowing you to use pipes and other shell features. `vim.system` only accepts a `string[]`, so we'll need to construct that ourselves.

stevearc (changes requested):
Made a couple of tweaks, mostly looks good! We just need to replicate the logic for running as a shell command.

lua/conform/runner.lua
lopi-py: I've changed the `build_cmd` function to always return `string[]` to run the shell command if required.

lua/conform/runner.lua
```diff
@@ -49,10 +52,13 @@ M.build_cmd = function(formatter_name, ctx, config)
       :gsub("$DIRNAME", ctx.dirname)
       :gsub("$RELATIVE_FILEPATH", compute_relative_filepath)
       :gsub("$EXTENSION", ctx.filename:match(".*(%..*)$") or "")
-    return command .. " " .. interpolated
+    return {
+      vim.split(vim.o.shell, " "),
+      vim.split(vim.o.shellcmdflag, " "),
+      string.format("{%s %s}", command, interpolated),
+    }
```
stevearc, line 59: I think this works in many cases, but isn't quite the same as what `jobstart` was doing before. I'd recommend checking out what's going on under the hood and trying to replicate that in lua. Some pointers:
- [jobstart builds the argument list](https://github.com/neovim/neovim/blob/27fb62988e922c2739035f477f93cc052a4fee1e/src/nvim/eval/funcs.c#L4248)
- [handle the case where the args are a string](https://github.com/neovim/neovim/blob/27fb62988e922c2739035f477f93cc052a4fee1e/src/nvim/eval.c#L5847-L5853)
- [function that builds the arg list](https://github.com/neovim/neovim/blob/27fb62988e922c2739035f477f93cc052a4fee1e/src/nvim/os/shell.c#L583-L606)

I expect this will be complicated enough to need its own utility function, with associated unit tests.

tests/runner_spec.lua
```diff
@@ -178,7 +194,15 @@ describe("runner", function()
...
-      assert.equal("echo " .. vim.fs.dirname(vim.api.nvim_buf_get_name(bufnr)) .. " | patch", cmd)
+      assert.are.same(
+        shell(
+          vim.fn.exepath("echo")
+            .. " "
+            .. vim.fs.dirname(vim.api.nvim_buf_get_name(bufnr))
+            .. " | patch"
+        ),
+        cmd
+      )
```
stevearc, line 205: ehhh...it's getting annoying to maintain these tests. They're pretty low-value, so I think we can just delete them. I'll do that and add a new test for a shell-command formatter.

stevearc (changes requested):

lua/conform/runner.lua
lopi-py: I realized I can just pull the function directly from C, so I don't think we need unit tests. Thanks for the guidance!

stevearc:
Great! I like the idea of calling the C function directly for consistency, but...I ended up replacing it with lua instead. I wanted to add a quick and dirty lua implementation for the case when ffi isn't available (LuaJIT isn't available for all platforms, so Neovim falls back to providing Lua 5.1). I briefly went over the C logic to make sure I was doing roughly the right thing, but it turned out to be so simple that I ended up just copying all the logic. And then once I had done that, it seemed kind of silly to have the two different implementations if we expect them to do the exact same thing, so I removed the ffi.

Thanks for putting the PR together! I'm excited to start migrating my plugins off of jobstart!

PR #528, opened 2024-09-02
Update deno_fmt.lua to allow esmodule extension

mashaal:
Adding the `.mjs` extension for Deno

https://docs.fileformat.com/web/mjs/


stevearc:
LGTM thanks!

PR #529, opened 2024-09-03
fix(log): prepend date to log lines

fei6409:
This helps people better distinguish whether a log line is fresh or aged.


stevearc:
Sure, seems fine to me

PR #534, opened 2024-09-12
feat: add kulala-fmt

gorillamoe:
kulala-fmt is an opinionated .http and .rest files linter and formatter.

x-ref: https://github.com/mistweaverco/kulala-fmt/issues/12


stevearc:
Thanks for the PR!

PR #549, opened 2024-10-17
feat: support python-ly

maxhambraeus:


lua/conform/formatters/python-ly.lua
```diff
@@ -0,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/frescobaldi/python-ly",
+    description = "A Python package and commandline tool to manipulate LilyPond files ",
+  },
+  command = "ly",
+  args = { "reformat", "$FILENAME" },
```
stevearc, line 8: You're passing in the filename, but the default value for `stdin` is true. Does this formatter work via stdin/stdout or does it operate on a file?

stevearc (changes requested):

lua/conform/formatters/python-ly.lua
maxhambraeus: My mistake, now it is correct

stevearc:
Thanks for the PR!

Issue #545, opened 2024-10-11
bug: prettierd config fails in monorepo (really specific)


wparsons-heins-globality:
### Neovim version (nvim -v)

NVIM v0.10.2

### Operating system/version

MacOS 14.5

### Read debugging tips

- [X] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

Log file: /Users/william.parsons-heins/.local/state/nvim/conform.log
          2024-10-11 15:56:31[INFO] Run prettierd on /Users/william.parsons-heins/[monorepo-name]/packages/[package-name]/src/client.js
          2024-10-11 15:56:31[DEBUG] Run command: { "/Users/william.parsons-heins/.local/share/nvim/mason/bin/prettierd", "/Users/william.parsons-heins/[monorepo-name]/packages/[package-name]/src/client.js" }
          2024-10-11 15:56:31[DEBUG] Run CWD: /Users/william.parsons-heins/[monorepo-name]/packages/[package-name]
          2024-10-11 15:56:31[INFO] prettierd exited with code 1
          2024-10-11 15:56:31[DEBUG] prettierd stdout: { "Error [ERR_MODULE_NOT_FOUND]: Cannot find package '[eslint-config]' imported from /Users/william.parsons-heins/[monorepo-name]/.prettierrc", "" }
          2024-10-11 15:56:31[DEBUG] prettierd stderr: { "" }
          2024-10-11 15:56:31[ERROR] Formatter 'prettierd' error: Error [ERR_MODULE_NOT_FOUND]: Cannot find package '[eslint-config]' imported from /Users/william.parsons-heins/[monorepo-name]/.prettierrc

### Describe the bug

I'm using yarn berry for a monorepo and I have prettier installed and configured at the top-level of the workspace, all sub-packages are formatted using this install and configuration. Using `nodeLinker: node-modules`, so there will be `node_modules` in each package and the root.

When I run the formatter in the terminal it is fine. But when I run conform formatting inside neovim on a sub-package file it will always fail.

[...]

PR #550, opened 2024-10-20
feat(cwd): add support for `contains_regex`

ro0gr:
solves #545, and #268(though it's closed, it has never really been solved).

I'd appreciate initial feedback if that's a good path to go for the project, so I can improve the solution.

todo:
 - [ ] cover `contains_regex` by tests
 - [ ] split unrelated tests(for pre-existing impl) into a separate PR
 - [ ] improve `markerFilesOrFn` types
 - [ ] final tests locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ tests/testfile.txt
 tests/fake_formatter_output
 scripts/nvim_doc_tools
 scripts/nvim-typecheck-action
+tmp/
--- a/lua/conform/formatters/prettierd.lua
+++ b/lua/conform/formatters/prettierd.lua
@@ -1,5 +1,40 @@
 local fs = require("conform.fs")
 local util = require("conform.util")
+
+local config_markers = {
+  ".prettierrc",
+  ".prettierrc.json",
+  ".prettierrc.yml",
+  ".prettierrc.yaml",
+  ".prettierrc.json5",
+  ".prettierrc.js",
+  ".prettierrc.cjs",
+  ".prettierrc.mjs",
+  ".prettierrc.toml",
+  "prettier.config.js",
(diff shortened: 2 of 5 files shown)


stevearc (changes requested):
This is adding a lot of custom handling and options to something that I think doesn't need to be this complicated. Is it likely that someone else will need a `contains_regex` feature inside the `util.root_file` function? The approach that I would rather see is just defining a custom cwd function here that does what you want.

```lua
cwd = function(self, ctx)
  return vim.fs.root(ctx.dirname, function(name, path)
    -- Do the simple contains check on the easy files here
    -- Then look for and parse the package.json
  end)
end
```

Given that this is also looking for a specific field in a json file, I'd prefer reading & parsing it rather than using regular expressions.

ro0gr:
Thanks for for your feedback. I agree it looks somewhat cumbersome to add this option for a single use case. My reason for this was to minimize amount of code to be copy-pasted if user wants to customize.

> Is it likely that someone else will need a contains_regex feature inside the util.root_file function?

There's also a `prettier` formatter that should follow the logic. This logic can be shared I think. Also there are more formatters that could rely on this pattern like eslint. But the difference with eslint binary is that it does nothing if no config was found, while prettier falls back to it's own defaults. So for prettier it is more important to get disabled on the conform side for such cases.

> Given that this is also looking for a specific field in a json file, I'd prefer reading & parsing it rather than using regular expressions.

Agree. If we internalze the lookup logic, it makes sense to treat it as json.

---

I'll work on an alternative pr based on you feedback.

ro0gr:
Close in favour of simplified version https://github.com/stevearc/conform.nvim/pull/564

PR #551, opened 2024-10-20
feat(formatters): add nufmt config

abhisheksingh0x558:


lua/conform/formatters/nufmt.lua
```diff
@@ -0,0 +1,8 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/nushell/nufmt",
+    description = "The nushell formatter",
+  },
+  command = "nufmt",
```
stevearc, line 7: Do you need to pass `--stdin` to get it to format from stdin/out? The docs seem to indicate that it's required.

stevearc (changes requested):
One question

lua/conform/formatters/nufmt.lua
abhisheksingh0x558: Thanks for pointing out. `$FILENAME` is what is required though.

stevearc: I would expect you to also need to add `stdin = false`

stevearc (changes requested):

lua/conform/formatters/nufmt.lua
stevearc: What I mean is, I would expect that either you are setting the option `stdin = false` and passing `args = { "$FILENAME" }` to indicate to the process that it should format a file, OR that you set `stdin = true` (or leave it out as that is the default) and pass `args = { "--stdin" }`. What I suspect is happening right now is that conform thinks it is running a formatter with stdin/stdout, but `nufmt` is formatting a file. That will work in simple cases, but in some situations cause very subtle bugs that are hard to track down.

abhisheksingh0x558: Got it! You are referring to conform option.
Fixed it!

stevearc:
Thanks for the PR!

PR #552, opened 2024-10-20
Add missing deno_fmt extensions

AndreasNasman:
I completed the `deno_fmt` extensions table with the [table of supported extensions in Deno's documentation](https://docs.deno.com/runtime/reference/cli/formatter/#supported-file-types). I added a couple of stable extensions that needed to be included.

Three unstable ones were also absent: Astro, Svelte, and Vue. These require the `--unstable-component` flag to work. If you don't provide the flag, the `deno fmt` command exists successfully (code 0) but does nothing.  Should the code here be refactored so that an info/warning/error message is logged if you format one of these files without the flag to improve debugging?

The [unstable-component flag documentation](https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component) mentions that Angular files are supported, but to my knowledge, there are no `.angular` files. Does anyone know what this means?

Also, can I remove `esmodule` from the `extensions` table? A `.mjs` file returns `javascript` as its extension, so I think `esmodule` isn't a valid extension in Neovim.

### DIFF
--- a/lua/conform/formatters/deno_fmt.lua
+++ b/lua/conform/formatters/deno_fmt.lua
@@ -1,6 +1,15 @@
+local conform = require("conform")
 local log = require("conform.log")
 
-local extensions = {
+-- Requires `--unstable-component` flag or
+-- `"unstable": ["fmt-component]` config option.
+-- https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component
+local unstable_extensions = {
+  astro = "astro",
+  svelte = "svelte",
+  vue = "vue",
+}
+local extensions = vim.tbl_extend("keep", {
   css = "css",
   esmodule = "mjs",
   html = "html",
@@ -15,12 +24,8 @@ local extensions = {
   typescript = "ts",
   typescriptreact = "tsx",
   yaml = "yml",
-  -- Requires `--unstable-component` flag or
-  -- `"unstable": ["fmt-component]` config option.
-  astro = "astro",
-  svelte = "svelte",
-  vue = "vue",
-}
+}, unstable_extensions)
+
 ---@type conform.FileFormatterConfig
 return {
   meta = {
@@ -31,10 +36,13 @@ return {
   args = function(self, ctx)
     local extension = extensions[vim.bo[ctx.buf].filetype]
 
-    -- TODO: How do I check if the user has passed in `--unstable-component` with e.g. `append_args` in their config?
-    print("self: " .. vim.inspect(self))
-    print("ctx: " .. vim.inspect(ctx))
-    if true then
+    if
+      vim.tbl_get(unstable_extensions, extension)
+      and not vim.list_contains(
+        vim.tbl_get(conform.formatters, "deno_fmt", "append_args") or {},
+        "--unstable-component"
+      )
+    then
       log.warn(
         "You are trying to format an unstable file type (."
           .. extension
--- a/lua/conform/formatters/deno_fmt.lua
+++ b/lua/conform/formatters/deno_fmt.lua
@@ -11,7 +11,6 @@ local unstable_extensions = {
 }
 local extensions = vim.tbl_extend("keep", {
   css = "css",
-  esmodule = "mjs",
   html = "html",
   javascript = "js",
   javascriptreact = "jsx",
--- a/lua/conform/formatters/deno_fmt.lua
+++ b/lua/conform/formatters/deno_fmt.lua
@@ -38,7 +38,7 @@ return {
       log.warn(
         "You are trying to format an unstable file type (."
           .. extension
-          .. ") without the corresponding `--unstable-component` flag. Add the flag to format the code. See the Deno documentation for more information: https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component"
+          .. ") without the corresponding `--unstable-component` flag. Add the flag to `append_args` to format your code. See the Deno documentation for more information: https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component"
       )
     end
 
--- a/lua/conform/formatters/deno_fmt.lua
+++ b/lua/conform/formatters/deno_fmt.lua
@@ -13,6 +13,11 @@ local extensions = {
   typescript = "ts",
   typescriptreact = "tsx",
   yaml = "yml",
+  -- Requires `--unstable-component` flag or
+  -- `"unstable": ["fmt-component]` config option.
+  astro = "astro",
+  svelte = "svelte",
+  vue = "vue",
 }
 ---@type conform.FileFormatterConfig
 return {
--- a/lua/conform/formatters/deno_fmt.lua
+++ b/lua/conform/formatters/deno_fmt.lua
@@ -43,16 +43,15 @@ return {
     }
 
     if
-      vim.tbl_get(unstable_extensions, extension)
+      unstable_extensions[extension]
       and not vim.list_contains(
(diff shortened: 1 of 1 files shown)


AndreasNasman:
I added a draft commit (ee7865c5546a81e3ae326df049e1d38b0e3c7a43) warning the user if they are trying to format an unstable file type with the `--unstable-component` flag. However, I need to figure out how to check if they've set the flag in their config. Does anyone know how to achieve that?

AndreasNasman:
> I added a draft commit ([a52d36d](https://github.com/stevearc/conform.nvim/commit/a52d36dba68b08a448f178c646a508fdd5dd7169)) warning the user if they are trying to format an unstable file type with the `--unstable-component` flag. However, I need to figure out how to check if they've set the flag in their config. Does anyone know how to achieve that?

I think I found a solution using `conform.formatters` and some `vim` table and list utilities. 👍

AndreasNasman:
I'm unsure how to check for the `"unstable": ["fmt-component]` config option in `deno.json`, so that's still missing.

lua/conform/formatters/deno_fmt.lua
```diff
@@ -16,11 +34,27 @@ return {
...
+        vim.tbl_get(conform.formatters, "deno_fmt", "append_args") or {},
+        "--unstable-component"
+      )
+    then
+      log.warn(
+        "You are trying to format an unstable file type (."
+          .. extension
+          .. ") without the corresponding `--unstable-component` flag. Add the flag to `append_args` to format your code. See the Deno documentation for more information: https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component"
+      )
+    end
```
stevearc, line 51: Could we just add `--unstable-component` automatically for the user here? If the formatter is running then that means they must have explicitly encoded _somewhere_ that they want to use deno fmt for that filetype. Are there any surprising risks to adding that flag?

stevearc (changes requested):
One question

lua/conform/formatters/deno_fmt.lua
AndreasNasman: That's a good point!

I don't think there are any risks in adding the flag. Even if the flag causes some problems, the formatting is the only thing that will suffer. The user will then know that `deno_fmt` causes the problem since they've chosen it as the formatter for the file they are trying to format. If they look it up, they will most likely come across this unstable flag, which explains by its name that everything might not be working 100 % yet.

The situation would be different if you could format the unstable files both with and without the flag, but in the current situation, you must include that flag. Otherwise, `deno fmt` succeeds, but no formatting takes place.

This would also fix this TODO:

> I'm unsure how to check for the `"unstable": ["fmt-component]` config option in `deno.json`, so that's still missing.

---

I could then change the MR so that the if statement adds `--unstable-component` to the return value and log it to `log.info` for debugging purposes. How does that sound?

lua/conform/formatters/deno_fmt.lua
```diff
@@ -16,11 +34,29 @@ return {
...
+      vim.tbl_get(unstable_extensions, extension)
+      and not vim.list_contains(
+        vim.tbl_get(conform.formatters, "deno_fmt", "append_args") or {},
+        "--unstable-component"
+      )
+    then
+      log.info(
+        "Adding `--unstable-component` to enable formatting of ."
+          .. extension
+          .. " files. See the Deno documentation for more information: https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component"
```
stevearc, line 55:
```suggestion
        "Adding `--unstable-component` to enable formatting of .%s files. See the Deno documentation for more information: https://docs.deno.com/runtime/reference/cli/formatter/#formatting-options-unstable-component", extension
```

lua/conform/formatters/deno_fmt.lua
```diff
@@ -16,11 +34,29 @@ return {
...
+    local formatter_args = {
       "fmt",
       "-",
       "--ext",
-      extensions[vim.bo[ctx.buf].filetype],
+      extension,
     }
+
+    if
+      vim.tbl_get(unstable_extensions, extension)
```
stevearc, line 46: No need to complicate things, we can just do
```suggestion
      unstable_extensions[extension]
```

lua/conform/formatters/deno_fmt.lua
```diff
@@ -16,11 +34,29 @@ return {
...
+      and not vim.list_contains(
+        vim.tbl_get(conform.formatters, "deno_fmt", "append_args") or {},
+        "--unstable-component"
+      )
```
stevearc, line 50: I don't love this reaching in to `conform.formatters` to check `append_args`. Will the command error if this argument is supplied more than once? Also, if the user needs fine-grained control over the arguments they can always override the `args` function themselves. Lastly, if we absolutely must, we could add another option to this formatter that allows the user to disable the auto-adding of this arg (but only if the first two options are insufficient for some reason).

stevearc (changes requested):
Couple nits

For the `esmodule` question: how does it currently behave when you run conform on a `.mjs` file? Presumably it'll be using the standard `.js` formatting; is that okay? If there is any difference between the two, then we may want to add another explicit check to see if `filetype == 'javascript' and vim.fn.fnamemodify(ctx.filename, ":e") == "mjs"` to set that as the formatter type. If not, then I guess we can remove it from the list.

lua/conform/formatters/deno_fmt.lua
AndreasNasman: Didn't know you could do that, thanks!

lua/conform/formatters/deno_fmt.lua
AndreasNasman: True, I was overthinking it 👍

AndreasNasman:
> For the `esmodule` question: how does it currently behave when you run conform on a `.mjs` file? Presumably it'll be using the standard `.js` formatting; is that okay? If there is any difference between the two, then we may want to add another explicit check to see if `filetype == 'javascript' and vim.fn.fnamemodify(ctx.filename, ":e") == "mjs"` to set that as the formatter type. If not, then I guess we can remove it from the list.

Yes, it uses the `.js` formatting. I cannot imagine why anyone would want to format them differently since they are structurally basically the same thing. The Deno formatting documentation also keeps `.js` and `.mjs` (and `.cjs` for that matter) under the same "JavaScript" umbrella term, so I think we can safely remove `module` from the extensions table.

lua/conform/formatters/deno_fmt.lua
AndreasNasman: > Will the command error if this argument is supplied more than once?

Unfortunately, yes. The formatter exists with an error saying that the same flag should only be used once:
image

But I agree; it feels rather clunky to reach into `append_args`. I think we could blatantly add `--unstable-component` since the error message indicates that this plugin already has the flag if you try to configure it yourself.

For context, we could skip the `if unstable_extensions[extension]` check since the `deno fmt` works fine if you provide the `--unstable-component` flag for _stable_ files (JavaScript, TypeScript, etc.), but I think it's better to add it only where needed (since it's unstable in some form).

> Also, if the user needs fine-grained control over the arguments they can always override the args function themselves. Lastly, if we absolutely must, we could add another option to this formatter that allows the user to disable the auto-adding of this arg (but only if the first two options are insufficient for some reason).

Good idea! 👍

AndreasNasman:
I added a new version with the `append_args` digging and `esmodule` removed.

stevearc:
Thanks for the PR!

PR #557, opened 2024-10-23
feat: add `biome-lint`

GWSzeto:
Adding support for using biome lint command. It is mainly used for the tailwind sorting (and more) which is why the `--unsafe` flag is there.


stevearc (changes requested):
The docs say
> Unsafe fixes may change the semantic of your program. Therefore, it’s advised to manually review the changes.

This isn't cause to ban the option altogether, but it makes me think it shouldn't be the default.

Also, is this formatter required? I see that `biome check` claims to run [the formatter and the linter](https://biomejs.dev/reference/cli/#biome-check), and it has the flags `--formatter-enabled` and `--linter-enabled` for enabling/disabling each individually.

GWSzeto:
you're right, just tried it with the `--unsafe` flag and got the intended results. Thanks!

PR #559, opened 2024-10-24
chore: add example layout python for direnv

lamchau:
added an example `.direnvrc` since i ran into this error

```
direnv: loading ~/dev/conform.nvim/.envrc
environment:860: python: command not found
direnv: Could not find python's version
./.envrc:2: pip: command not found
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,3 +1,3 @@
 export VIRTUAL_ENV=venv
-layout python
-python -c 'import pyparsing' 2> /dev/null || pip install -r scripts/requirements.txt
+layout python python3
+python -c 'import pyparsing' 2>/dev/null || pip install -r scripts/requirements.txt


stevearc (changes requested):
Python is a [built-in layout](https://github.com/direnv/direnv/wiki/Python) for direnv. Are you using an old version of direnv that doesn't have this?

lamchau:
hmm, pretty up to date but til it's supported by default! it's installed via `homebrew` for me maybe that's the difference?

```sh
$ direnv --version
2.35.0
```

```sh
$ direnv allow
direnv: loading ~/dev/conform.nvim/.envrc
environment:860: python: command not found
direnv: Could not find python's version
./.envrc:2: pip: command not found
```

lamchau:
it turns out i needed to use `layout python python3` for it to work

stevearc:
If you want to change the .envrc to use `layout python3` that sounds fine to me

lamchau:
up to you! pushed up a change that drops the .direnvrc and updates the layout. if it works we can merge it in otherwise, i can just keep that change locally on my end :)

PR #562, opened 2024-10-31
feat: add support for nginx-config-formatter

dangdd2003:
I saw nginx is very popular web server and is used widely, but there aren't any support for its config file now, so i add it. The config file without formatting is definitely a mess.

The doc is [here](https://github.com/slomkowski/nginx-config-formatter).


stevearc:
Thanks for the PR!

PR #564, opened 2024-11-01
fix(prettierd): respect `prettier` config field in the "package.json"

ro0gr:
solves #545, and #268


lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,55 @@
...
+    f:close()
+
+    return vim.fn.json_decode(file_content)
+  else
+    print("Error: Unable to open file " .. file)
+    return nil
+  end
+end
+
+-- TODO: share this with "lua/conform/formatters/prettier.lua"
```
ro0gr, line 34: I'd work on it in a separate pr, if this pr is accepted.

lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,55 @@
...
+}
+
+local function read_json(file)
+  local f = io.open(file, "r")
+
+  if f then
+    local file_content = f:read("*all") -- Read entire file contents
+    f:close()
+
+    return vim.fn.json_decode(file_content)
```
stevearc, line 27: Let's use vim.json.decode here

lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,55 @@
...
+  ".prettierrc.js",
+  ".prettierrc.cjs",
+  ".prettierrc.mjs",
+  ".prettierrc.toml",
+  "prettier.config.js",
+  "prettier.config.cjs",
+  "prettier.config.mjs",
+}
+
+local function read_json(file)
```
stevearc, line 20: Let's add some type annotations
```suggestion
---@param file string
---@return nil|table
local function read_json(file)
```

lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,55 @@
...
+  local f = io.open(file, "r")
+
+  if f then
+    local file_content = f:read("*all") -- Read entire file contents
+    f:close()
+
+    return vim.fn.json_decode(file_content)
+  else
+    print("Error: Unable to open file " .. file)
```
stevearc, line 29: We have a logger for things like this
```suggestion
    log.warn("Unable to open file %s", file)
```

tests/formatters/pretierd_spec.lua
```diff
@@ -0,0 +1,127 @@
+require("plenary.async").tests.add_to_env()
```
stevearc, line 1: filename has typo

stevearc (changes requested):
Couple of nits

lua/conform/formatters/prettierd.lua
ro0gr: I feel it's more applicable to throw an exception here. I'll check it.

stevearc: I disagree. There could be some people that use json5 for their package.json file (uncommon, but I've seen it -_-). That would cause this formatter to fail and spit out an error every time. It would be better to instead fall back to a best-effort guess using the other marker files as that will work just fine in most cases.
I could be convinced that it should log an error instead of a warning though, just to make sure it's visible by default in `:ConformInfo`

ro0gr: Oh.. right. Then I think a warning should be fine until we are able to read json5.

lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,66 @@
...
+  "prettier.config.cjs",
+  "prettier.config.mjs",
+}
+
+---@param file string
+---@return nil|table
+local function read_json(file)
+  local f = io.open(file, "r")
+  if not f then
+    error("Unable to open file " .. file)
```
ro0gr, line 25: So it turns out that's what I meant when I said

> I feel it's more applicable to throw an exception here. I'll check it.

I haven't tested this specific code path cause the `read_json` is an internal util, and we should never reach this line until we don't have bugs in the `package.json` logic.

lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,66 @@
...
+  local file_content = f:read("*all") -- Read entire file contents
+  f:close()
+
+  local ok, json = pcall(function()
+    return vim.json.decode(file_content)
+  end)
+
+  if not ok then
+    local log = require("conform.log")
+    log.error("Unable to parse json file " .. file)
```
ro0gr, line 37: If you prefer the `warn` here, just let me know please

lua/conform/formatters/prettierd.lua
```diff
@@ -1,5 +1,66 @@
...
+  ".prettierrc.yaml",
+  ".prettierrc.json5",
+  ".prettierrc.js",
+  ".prettierrc.cjs",
+  ".prettierrc.mjs",
+  ".prettierrc.toml",
+  "prettier.config.js",
+  "prettier.config.cjs",
+  "prettier.config.mjs",
+}
```
pmanole-volvocars, line 18: No issue from me, but I just wanted to point out the fix I proposed for prettier (not prettierd) in #567 - specifically, these formatters should not run on ALL projects, meaning that if package.json is detected, they run. Some projects have other formatters and keeping package.json here will trigger prettier alongside their preferred formatter, leading to linting errors and screen flicker because of the different formatting styles being applied one after another.

ro0gr: Yes, treating package.json existence as a sufficient condition for the prettierd being enabled is not correct. Ignoring package.json is not correct as well. The point of the PR is to fix the package.json treatment. And yes, I'd like to share this logic with the prettier formatter if this PR finds it's way to the master.

stevearc:
Thanks for the PR!

PR #567, opened 2024-11-05
fix(prettier): update prettier formatter configuration to only trigger on projects with a .prettierrc file

pmanole-volvocars:
This change updates the configuration of the Prettier formatter to prevent it from running based solely on the presence of a package.json file. Going forward, Prettier will only activate if a .prettierrc-type configuration file is present.

### Rationale
By limiting Prettier activation to projects with a dedicated Prettier configuration file, this change respects projects that use alternative formatters like dprint or eslint with custom plugins, ensuring that these tools can function without Prettier inadvertently overriding their formatting.

For example, templates such as create-effect-app, which use dprint and don’t generate a .prettierrc file, were previously impacted by Prettier, which introduced formatting inconsistencies (e.g., trailing commas) that conflicted with the project's style preferences and triggered linter errors.

### Impact
This update allows developers to use their preferred formatter consistently. In projects without a .prettierrc file, Prettier will no longer apply formatting, avoiding unnecessary changes and ensuring that the selected formatter is the sole tool enforcing code style.

### DIFF
--- a/lua/conform/formatters/prettier.lua
+++ b/lua/conform/formatters/prettier.lua
@@ -72,6 +72,5 @@ return {
     ".prettierrc.toml",
     "prettier.config.js",
     "prettier.config.cjs",
-    "package.json",
   }),
 }


pmanole-volvocars:
Fixed merge conflict (178bcda765fdccd2dab07974a75be3f6576b8a96).

pmanole-volvocars:
update: asked Chat GPT to rewrite and improve PR title and description

NullVoxPopuli:
This is a good change. 

The package.json should not be used _unless_ it declares a prettier key/object, and prettier is installed.

I don't think conform checks those things currently, so it's best to remove package.json

stevearc:
I would prefer to share the logic in #564 with prettier instead of removing `package.json` entirely

PR #568, opened 2024-11-06
Add support for Clojure standard-clj

sebastian-palma:
Adding support for [`standard-clj`](https://github.com/oakmac/standard-clojure-style-js).

The filename differs from the other standard linters, but wanted to preserve the command name.

(thanks for conform btw 🍻)


stevearc:
Thanks for the PR!

PR #570, opened 2024-11-11
fix(prettier): respect prettier config field in "package.json"

ro0gr:
continues https://github.com/stevearc/conform.nvim/pull/564
addresses https://github.com/stevearc/conform.nvim/pull/567

### DIFF
--- a/lua/conform/util.lua
+++ b/lua/conform/util.lua
@@ -1,3 +1,5 @@
+local log = require("conform.log")
+
 local M = {}
 
 ---Find a command in node_modules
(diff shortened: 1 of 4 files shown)


lua/conform/util.lua
```diff
@@ -246,4 +248,60 @@ M.shell_build_argv = function(cmd)
...
+    return nil
+  end
+
+  return json
+end
+
+-- @param self conform.FormatterConfig
+-- @param ctx conform.Context|conform.RangeContext
+-- @return string[]|nil args the arguments for setting a `prettier` parser if they exist in the options, nil otherwise
+M.prettier_cwd = function(self, ctx)
```
ro0gr, line 275: I have a gut feeling it should be a part of some external dependency, so it could be shared between other plugins like nvim-lint, etc. Until then this is only used by 2 formatters, so this does not seem to be the best to share it via a general utils file that is used everywhere. I'm considering to move this into the `lua/conform/util/prettier-cwd` or smth like that now.

tests/util/prettier_spec.lua
ro0gr, line 1: As we share the logic between 2 formatters I'd avoid duplicating tests for the second formatter, so I'd test the `prettier_cwd(` util instead

stevearc (changes requested):
I think this could be a lot simpler. In general the `util` functions are general enough that they might be shared by a lot of different formatters. What if for the prettier formatter you just did

```lua
cwd = require("conform.formatters.prettierd").cwd
```

ro0gr:
> What if for the prettier formatter you just did

TLDR; OK, this is definitely easier to do, so I'll stop overthinking it and make a step back. Thanks.

I've tried it initially, but then I could not decide which of both formatters should really own the cwd logic. IMO neither of them should, cause in my "ideal world"_(it's beyond the scope of this project, I know!)_ this logic could be shared out of the conform, for instance with nvim-lint. That's why I feel keeping this as a shared util(via `util/prettier.lua` is a bit closer to my "ideal world".

ro0gr:
close in favour of https://github.com/stevearc/conform.nvim/pull/573

Issue #571, opened 2024-11-12
bug: xmlformatter has inconsitent name


Skyppex:
### Neovim version (nvim -v)

NVIM v0.10.2

### Operating system/version

Windows 11

### Read debugging tips

- [X] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

I have no log file to provide

### Describe the bug

In mason.nvim, on the command line, and also the code for this plugin, the xml formatter has the name `xmlformatter`, but when setting up the formatter you must use the name `xmlformat`. This is undocumented and i found myself digging in the code to learn about it.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

```lua
formatters_by_ft = {
    xml = "xmlformat", -- this works
},
```

### Expected Behavior

The name of the xml formatter should be `xmlformatter`.

```lua
formatters_by_ft = {
    xml = "xmlformatter",
},
```

### Additional context

I don't believe this requires an examle configuration. The problem as i see it is that the `formatters/xmlformat.lua` file should be renamed to `formatters/xmlformatter.lua`.

PR #572, opened 2024-11-12
fix(names): change name of xml formatter

Skyppex:
# Changes

- rename: formatters/xmlformat.lua -> formatters/xmlformatter.lua

closes: #571


Skyppex:
As a note, i was reading the name wrong for the cli program name. That is actually `xmlformat`, but i still think it should be called `xmlformatter` since thats consitent with other popular plugins in the nvim ecosystem (mason.nvim)

stevearc (changes requested):
This is going to break people's configs. Please add a shim like we did in https://github.com/stevearc/conform.nvim/blob/e3263eabbfc1bdbc5b6a60ba8431b64e8dca0a79/lua/conform/formatters/ruff.lua so that `xmlformat` continues to work

stevearc:
That's just because a new version of LuaLS released and added some strictness. I'll fix it on master.

Skyppex:
> That's just because a new version of LuaLS released and added some strictness. I'll fix it on master.

I see, thanks for merging :)
I'm not very versed in the lua ecosystem yet xD

Issue #582, opened 2024-11-22
feature request: provide feedback if unknown formatter is referenced


ro0gr:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

Currently if user has a typo in a formatter name or reference a non-existing, there is no reasonable feedback about it. 

Assuming the following config:
```
      formatters_by_ft = {
        javascript = { 'eslintd', 'prettierd' },
      },
```

I have the following `ConformInfo` output(ignoring the "Log file" output cause there are no any related entries):

<img width="594" alt="Знімок екрана 2024-11-22 о 14 02 35" src="https://github.com/user-attachments/assets/e4a8a1c1-c19d-4678-b93d-39f9ca844dd1">

The issue is that the proper name should use an underscore - `eslint_d`. But that was not obvious for me and `eslint` turned out to be disabled for for quite a while. 

On the screenshot you can see two `unavailable` entries:
 - The `prettierd` warning is intentional cause it's a known formatter which is enabled for the filetype, but it's just not applicable for the current file due to a missing prettier setup in the project. 
 - The `eslintd` case is different cause the formatter does not even exist in Conform 🚨 
 
 I believe UX could win if we made a more clear distinction between these 2 types of formatter results. 

Maybe unkown formatters should not be placed along with the known formatters, so removing it from the `Formatters for this buffer:` section, and place it in a separate category(e.g. `Failed to inititialze:`). 

If it's placed there intentionally, maybe we could improve the messsage?

The exisitng message didn't give me a clue on what's wrong:

> Formatter config missing or incomplete

which config does it mean? eslint, conform? Also it's colored the same as a normal `unavailable` entries. 

Happy to improve it if it's desired.

### What is the significance of this feature?

nice to have

ro0gr:
OK, I think highlighting it as an error, and maybe improving the message should make me happy :) 

Smth like:
<img width="1002" alt="Знімок екрана 2024-11-22 о 16 52 29" src="https://github.com/user-attachments/assets/32bdd251-c36c-4466-b716-689ce5916683">

Any objections?

stevearc:
SGTM. Will require a bit of refactoring because right now all the availability checks are the same severity

PR #583, opened 2024-11-22
improve error message for unknown formatters

ro0gr:
This tries to solve the #582. 

I'm not really sure if it makes the feedback noticeably better. But I think with this kind of message I might have noticed my configuration issue earlier.

I'd say it's still missing highlighting the `unavailable` keyword with an error color, cause this kind of error is probably caused by the misconfiguration on the user side, and it makes sense to make it as clear as possible. WDYT?


stevearc:
This change LGTM. Did you want me to merge or did you want to try to change the highlighting?

ro0gr:
Thanks. Yeah, I'll give a try to highlighting improvements.

ro0gr:
It's ready for review now

stevearc:
LGTM thanks for the PR!

PR #589, opened 2024-11-28
Rename awk formatter to gawk

fredrikfoss:
I made a similar pull request in mfussenger/nvim-lint ([#684]) a while back.
Same as in that case, gawk is more or less the only awk implementation that
supports formatting of awk files, thus I find it make sense to have the
formatter named gawk instead. Could also prevent problems for people who use
Alpine as an example, which by default uses Busybox's awk implementation, but
still would like to have an awk formatter configured for when gawk is available.

[#684]: https://github.com/mfussenegger/nvim-lint/pull/684


stevearc (changes requested):
Can you leave the current formatter in place and flag it as deprecated like we do for typstfmt? https://github.com/stevearc/conform.nvim/blob/master/lua/conform/formatters/typstfmt.lua

This will remove it from the documentation while preserving backwards compatibility

stevearc:
Great, thanks for the PR!

PR #590, opened 2024-12-02
fix(docs): link to `trim_newlines` formatter

nelson137:


stevearc (changes requested):
These docs are actually autogenerated, you'll need to update the link in the `trim_newlines.lua` file

nelson137:
> These docs are actually autogenerated, you'll need to update the link in the `trim_newlines.lua` file

Ah gotcha, I can do that.

nelson137:
@stevearc done :)

stevearc:
Thanks!

PR #591, opened 2024-12-02
fix(runner): replace `$FILENAME` last, since it can contain `%` which would fail when `gsub` `$RANDOM`

folke:
If the `basename` contains `%` characters, that line fails.
This fixes that.

Fixes https://github.com/folke/snacks.nvim/issues/187


stevearc:
LGTM thanks!

Ocent22:
This should be an awesome logical fast means to scale through.

PR #595, opened 2024-12-10
feat(typst): add `prettypst` formatter

dustypomerleau:
Thanks so much for all the time you put into conform!

[`prettypst`](https://github.com/antonWetzel/prettypst) is the most configurable formatter currently available for Typst, and provides another option to users migrating away from the deprecated version of `typstfmt`.


stevearc:
Thanks for the PR!

PR #596, opened 2024-12-11
feat: add equalprg

### LINKED ISSUE #555: feature request: a new formatter for Vim's own C-indenting

### Describe the feature

Vim and Neovim have built-in functionality for indenting C-style programs:  
https://neovim.io/doc/user/indent.html#_1.-indenting-c-style-programs

It would be awesome to be able to use this as a formatter like any other in Conform.

### Provide background

All of the common C formatters have too many configuration options. As someone who only dabbles with C on occasion, Vim's built-in C-indenting is more than enough for my needs. But I'd love to have it trigger on save rather than running it manually. And having it set up as a formatter in Conform would mean not needing to duplicate format-on-save logic.

nice to have

### Additional details

_No response_

heygarrett:
Closes #555 

Vim and Neovim have built-in formatting functionality:  
- `:help equalprg`
- `:help C-indenting`

This PR adds equalprg as a formatter which does the following:
1. Store current cursor position in a variable
2. `gg`
3. `=G`
4. Restore cursor position from variable


stevearc (changes requested):
This may appear to work in the simple case, but it will cause errors as soon as you try to run it with more than one formatter. Lua function formatters need to

1. not modify the buffer
2. pass the formatted text back as an argument to the callback

heygarrett:
Ah, I didn't realize the buffer can't be modified. Are there any existing formatters that use a Lua function but act directly on text in a buffer that I could look at as an example? I wonder if using a temporary file would work. 🤔

stevearc:
There aren't any other formatters that do something like that right now. I agree that a temporary buffer is probably the way to go for this, but it'll likely be a bit tricky. Usually the FileType autocmd will set the `indentexpr` or `equalprg` on the buffer, but if you try to set the filetype on a temp buffer it won't work as expected during format_on_save or format_after_save because we're _already_ in an autocmd. You'll likely need to manually copy over the relevant option values to the temp buffer.

PR #598, opened 2024-12-15
feat: add gojq

telemachus:


telemachus:
This may be a niche request, and it's easy enough already to add it as a custom formatter. But I figured I would share it anyhow. I may not be the only Gopher who will appreciate the option.

Issue #585, opened 2024-11-25
feature request: Organize import using Biome.js for typescript, javascript, typescriptreact, javascriptreact


rahulpeacock:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

I want to organize the imports using `biome.js` when I format my buffer/file using `biome.js` for typescript, javascript, typescriptreact, javascriptreact

### What is the significance of this feature?

strongly desired

### Additional details

For now I am using the `args` to organize the imports. But it is highly recommend to have a config to organize the imports using biome

This is a workaround for now
```lua
-- confirm.lua
return {
  {
    "stevearc/conform.nvim",
    opts = {
      formatters_by_ft = {
        lua = { "stylua" },
        ["javascript"] = { "biome" },
        ["javascriptreact"] = { "biome" },
        ["typescript"] = { "biome" },
        ["typescriptreact"] = { "biome" },
        ["json"] = { "biome" },
        ["css"] = { "biome" },
      },
      formatters = {
        biome = {
          command = "biome",
          args = { "check", "--formatter-enabled=true", "--linter-enabled=false", "--organize-imports-enabled=true", "--write", "--stdin-file-path", "$FILENAME" },
        },
      },
    },
  },
}
```

stevearc:
According [to the docs](https://biomejs.dev/guides/configure-biome/) it looks like the formatter, linter, and import sorter are all enabled by default. The default conform `biome` formatter calls `biome format`. If you want to use `biome check`, use the `biome-check` formatter.

Also, to make changes to a formatter args, you can provide `append_args` or `prepend_args` instead of overriding the args entirely.

rahulpeacock:
@stevearc Yes I can pass args but I can't pass the arg `--import-sort-enable` to the `biome format` command because it doesn't support that arg

rahulpeacock:
@stevearc can you share any config with `biome-check`.

andrew-t-james-core:
@rahulpeacock I had some luck with this [setup](https://github.com/LazyVim/LazyVim/issues/2116#issuecomment-2489562068).

rahulpeacock:
@andrew-t-james-core this is also a workaround but is there a default config from `conform` ??

stevearc:
Can you not just use the built-in `biome-check` formatter? It runs `biome check` which, from my understanding, will organize the imports
```lua
      formatters_by_ft = {
        lua = { "stylua" },
[...]

PR #599, opened 2024-12-17
feat: biome organize imports

igorlfs:
Hello!

This PR adds another biome formatter, to **only** organize imports (in a similar manner to ruff, with ruff's [organize-imports](https://github.com/stevearc/conform.nvim/blob/master/lua/conform/formatters/ruff_organize_imports.lua)).

The major reason for another formatter is that there are some [known limitations](https://biomejs.dev/internals/language-support/#html-super-languages-support) when using biome's regular formatter with HTML super languages, such as Svelte and Vue.

### DIFF
--- a/lua/conform/formatters/biome-organize-imports.lua
+++ b/lua/conform/formatters/biome-organize-imports.lua
@@ -7,7 +7,15 @@ return {
   },
   command = util.from_node_modules("biome"),
   stdin = true,
-  args = { "check", "--write", "--formatter-enabled=false", "--stdin-file-path", "$FILENAME" },
+  args = {
+    "check",
+    "--write",
+    "--formatter-enabled=false",
+    "--linter-enabled=false",
+    "--organize-imports-enabled=true",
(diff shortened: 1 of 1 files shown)


igorlfs:
Uh, I just realized I can achieve the same result by disabling formatting, inside Biome's configuration. Specifically, here's the snippet for my use case (Svelte):

```json
"formatter": {
        "enabled": true,
        "ignore": [
            "**/*.svelte",
        ]
    }
```

Coupled with the `biome-check` formatter, this **only** organizes imports. 

 I was considering closing this PR, but I noticed it was mentioned elsewhere, and it would close https://github.com/stevearc/conform.nvim/issues/585 (I guess)

cc @andrew-t-james-core @rahulpeacock

stevearc:
How would you expect this formatter to be used, and is the use case different from the `biome-check` formatter?

stevearc:
Sounds reasonable to me. Thanks for the PR!

PR #605, opened 2024-12-24
fix: create ConformInfo regardless of setup

saccarosium:
There is little to no benefit to create `ConformInfo` only if the user calls the setup function since you can use this plugin without never calling it.


stevearc:
LGTM thanks!

PR #608, opened 2024-12-25
Small nitpipicks

saccarosium:
Hi,
this are some small nitpicks that IMHO make the code a bit more clear.

- You can simply pass the functions to `vim.schedule_wrap`
- Why do we need `allowed_default_*_opts` globally when we are using them only in one function?
- You can force the lsp to cast a value instead of creating another function
- Use `tbl_isempty` instead of checking the legth with `#`

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -14,25 +14,21 @@ M.default_format_opts = {}
 
 -- Defer notifications because nvim-notify can throw errors if called immediately
 -- in some contexts (e.g. inside statusline function)
-local notify = vim.schedule_wrap(function(...)
-  vim.notify(...)
-end)
-local notify_once = vim.schedule_wrap(function(...)
-  vim.notify_once(...)
-end)
+local notify = vim.schedule_wrap(vim.notify)
+local notify_once = vim.schedule_wrap(vim.notify_once)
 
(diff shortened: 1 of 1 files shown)


stevearc:
I understand that this PR is likely coming from a place of using the plugin, liking it, and wanting to contribute. I want you to know that that is a good instinct and I appreciate the thought behind it. With that said, this is not a useful PR and I would generally advise you against making these types of PRs. Every PR takes some amount of my time and energy as a maintainer, and what I hope to get out of it is that the benefit to the project is worth that input. The benefits to a nitpick PR are very small, subjective, and often a matter of taste. 

Another reason why I would generally advise against this is that you're unfamiliar with the project and something that seems oddly verbose for no reason might actually have a reason. For example, we're not just calling `vim.schedule_wrap(vim.notify)` because `vim.notify` can get patched to be another function, and we want to have the latest version of `vim.notify` not the version that existed when this file was required. Some of the other nits you changed have similar, subtle reasons for why they are as they are.

So I will not be merging this, but if you want to contribute some other way in the future, say by adding a formatter or fixing a bug, I would love to look at another PR!

saccarosium:
Ok make sense I will avoid this kind of PRs in the future. Thank you for the kindness :).

PR #611, opened 2024-12-29
feat: add pycln

Quitlox:


lua/conform/formatters/pycln.lua
```diff
@@ -0,0 +1,18 @@
...
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/hadialqattan/pycln",
+    description = "A Python formatter for finding and removing unused import statements.",
+  },
+  command = "pycln",
+  args = {
+    "--silence",
+    "-a",
```
stevearc, line 11: According to the docs the default behavior tries to avoid removing imports that could have side effects, and providing this `-a` will remove even those imports. I think that for a default configuration it makes more sense to omit the `-a`. I like the defaults to avoid potentially logic-changing formats, and it's easier to add the `-a` back in your own config using `prepend_args` than it is to remove `-a` from the list of arguments.

stevearc (changes requested):

lua/conform/formatters/pycln.lua
Quitlox: Hmm, the docs do specify that the recommended way is to run it using `-a`, which is my experience as well (see: https://hadialqattan.github.io/pycln/#/?id=the-simplest-usage). Without specifying `-a` it will leave many "third party" imports (which is most in any decently sized project) in, as it cannot decide whether they have side effects. In practice, it also seems easier and less annoying to sometimes have to add a `#nopycln: import` in case of overeagerness, instead of the tool leaving behind many unused third party packages.

I get that adding the `-a` flag is easier than removing it, but for a best out-of-the-box experience I would leave it in.

Decision is ultimately yours though :)

(Apologies for the slow response, the notification was lost on me.)

stevearc: Yeah, I think I'd prefer leaving it out by default. Similar to how the tool itself does not make `-a` the default but instead requires you to explicitly opt-in. They may recommend it, but they don't do it for you by default. I think we should do the same.

Quitlox: Alright, I removed it.

PR #614, opened 2024-12-31
fix(format-queries): require nvim-treesitter before scaning rtp

phanen:
`scripts/format-queries.lua` live in https://github.com/nvim-treesitter/nvim-treesitter repository.
Those who lazy load this plugin should require it before searching the `rtp`.


stevearc:
LGTM thanks!

PR #616, opened 2024-12-31
fix: Add `--respect-ignores` to the stylua command

### LINKED ISSUE #615: bug: stylua doesn't respect `.styluaignore` file by default after v2.0.0
### Neovim version (nvim -v)

`NVIM v0.10.3 Build type: Release LuaJIT 2.1.1716656478`

### Operating system/version

Gentoo Linux 2.17

### Read debugging tips

- [X] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

### Log file

---

### Describe the bug

Since [v2.0.0](https://github.com/JohnnyMorganz/StyLua/releases/tag/v2.0.0) StyLua adds the `--respect-ignores` flag ([PR](https://github.com/JohnnyMorganz/StyLua/pull/912)) and changes the default behaviour of `--stdin-filepath`:

> Now, as we approach v2, we make the breaking change to make `--stdin-filepath` not respect ignores by default. Pass `--respect-ignores` to preserve original behaviour.

As result, conform is applying formatting to files in the `.styluaignore` file.

breaking (some functionality is broken)

### Steps To Reproduce

1. In a new dir write the `repro.lua`
2. Run `nvim -u repro.lua`
3. Use `:SetCase`
4. `:w` to write the `.styluaignore` file into the cwd
5. `<C-w>j` and `:w` to write the `foo.lua` file.
6. `foo.lua` is being formatted.

### Expected Behavior

Since `foo.lua` is in the `.styluaignore` file, it shouldn't be formatted.

### Minimal example file

Check the minimal init.lua `SetCase` command.

### Minimal init.lua

```Lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    "stevearc/conform.nvim",
    dependencies = {
      {
        "williamboman/mason.nvim",
        build = { ":MasonUpdate" },
        config = function(_, opts)
          require("mason").setup(opts)
          local mr = require("mason-registry")
          if not mr.is_installed("stylua") then
            local p = mr.get_package("stylua")
            p:install()
          end
        end,
      },
    },
    opts = {
      formatters_by_ft = { lua = { "stylua" } },
      format_on_save = { timeout_ms = 500, lsp_fallback = true },
      -- FIX: Uncomment to fix the issue
      -- formatters = { stylua = { prepend_args = { "--respect-ignores" } } },
    },
  },
})

vim.api.nvim_create_user_command("SetCase", function()
  vim.cmd("edit foo.lua")
  vim.api.nvim_buf_set_lines(0, 0, -1, false, { 'return { foo = function() return { foo = 123, bar = "buz", } end }' })

  vim.cmd("split .styluaignore")
  vim.api.nvim_buf_set_lines(0, 0, -1, false, { "foo.lua" })
  vim.notify("Done. Save `.styluaignore` and then `foo.lua`. `foo.lua` will be formatted")
end, {})

vim.notify("Use `:SetCase` to fill the case content")
```

### Additional context

To preserve the previous behaviour, `--respect-ignores` should be added in the args of the `lua/conform/formatters/stylua.lua` module.

Edit: Check the PR #616

polirritmico:
Fixes #615

### DIFF
--- a/lua/conform/formatters/stylua.lua
+++ b/lua/conform/formatters/stylua.lua
@@ -6,7 +6,7 @@ return {
     description = "An opinionated code formatter for Lua.",
   },
   command = "stylua",
-    args = {
+  args = {
     "--search-parent-directories",
     "--respect-ignores",
     "--stdin-filepath",
--- a/lua/conform/formatters/stylua.lua
+++ b/lua/conform/formatters/stylua.lua
@@ -6,7 +6,13 @@ return {
     description = "An opinionated code formatter for Lua.",
   },
   command = "stylua",
-  args = { "--search-parent-directories", "--stdin-filepath", "$FILENAME", "-" },
+    args = {
+    "--search-parent-directories",
+    "--respect-ignores",
+    "--stdin-filepath",
+    "$FILENAME",
+    "-",
+  },
   range_args = function(self, ctx)
     local start_offset, end_offset = util.get_offsets_from_range(ctx.buf, ctx.range)
     return {


stevearc:
Thanks for the PR!

hgkjshegfskef:
After this change, I get an error in the LazyVim distribution: `Formatter 'stylua' error: error: Found argument '--respect-ignores' which wasn't expected, or isn't valid in this context`. Not sure if this is upstream- or downstream-specific. I opened a ticket for this in downstream: https://github.com/LazyVim/LazyVim/issues/5375

hgkjshegfskef:
Nevermind, I discovered that my version of StyLua was 1.5 years outdated and updating it. The issue does no longer occur. That is, I simply didn't have StyLua >= 2.0.0, where this option was introduced.

PR #623, opened 2025-01-14
feat: add pre and post formatter hooks

itmecho:
This is a feature I've added to my local copy of `conform.nvim` which I wondered if you would be interested in adding! I would've created and issue instead but I thought I'd just do a PR instead as I had already made the changes locally.

## What?
Auto command triggers for `ConformPreFormatter` and `ConformPostFormatter` which fire before and after each formatter that is called. This allows hooking into the format process for each formatter.

## Why?
I've added this as the formatters I have to run for work can sometimes take a while and there doesn't seem to be any feedback as to which formatter is currently running. 

Adding these hooks enables me to have the following auto commands:
```lua
vim.api.nvim_create_autocmd("User", {
  pattern = 'ConformPreFormatter',
  callback = function(event)
    print('running ' .. event.data.formatter.name)
  end
})

vim.api.nvim_create_autocmd("User", {
  pattern = 'ConformPostFormatter',
  callback = function(event)
    print(string.format('finished running %s: ok=%s', event.data.formatter.name, event.data.ok))
  end
})
```

which results in the following:

https://github.com/user-attachments/assets/1f9206f5-2598-43d0-8c75-c8bc79c0eea8

### DIFF
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -298,9 +298,14 @@ local last_run_errored = {}
 ---@param callback fun(err?: conform.Error, output?: string[])
 ---@return integer? job_id
 local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, callback)
+  local autocmd_data = {
+    formatter = {
+      name = formatter.name,
+    },
+  }
   vim.api.nvim_exec_autocmds("User", {
     pattern = "ConformPreFormatter",
-    data = { formatter = formatter },
+    data = autocmd_data,
   })
   log.info("Run %s on %s", formatter.name, vim.api.nvim_buf_get_name(bufnr))
   log.trace("Input lines: %s", input_lines)
@@ -313,6 +318,11 @@ local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, c
     else
       last_run_errored[formatter.name] = false
     end
+    autocmd_data["err"] = err
+    vim.api.nvim_exec_autocmds("User", {
+      pattern = "ConformPostFormatter",
+      data = autocmd_data,
+    })
   end)
   if config.format then
     local err_string_cb = function(err, ...)
@@ -438,13 +448,6 @@ local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, c
           })
         end
       end
-      vim.api.nvim_exec_autocmds("User", {
-        pattern = "ConformPostFormatter",
-        data = {
-          formatter = formatter,
-          success = result.code == 0,
-        },
-      })
     end)
   )
   if not ok then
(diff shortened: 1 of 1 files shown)


lua/conform/runner.lua
```diff
@@ -298,6 +298,10 @@ local last_run_errored = {}
 ---@param callback fun(err?: conform.Error, output?: string[])
 ---@return integer? job_id
 local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, callback)
+  vim.api.nvim_exec_autocmds("User", {
+    pattern = "ConformPreFormatter",
+    data = { formatter = formatter },
+  })
   log.info("Run %s on %s", formatter.name, vim.api.nvim_buf_get_name(bufnr))
   log.trace("Input lines: %s", input_lines)
   callback = util.wrap_callback(callback, function(err)
```
stevearc, line 307: We should try to provide the guarantee that if `ConformPreFormatter` is called, then `ConformPostFormatter` will also be called. Right now there are some error paths that won't hit the second autocmd. I believe if we move it into this `util.wrap_callback` function it will always get called.

lua/conform/runner.lua
```diff
@@ -298,6 +298,10 @@ local last_run_errored = {}
 ---@param callback fun(err?: conform.Error, output?: string[])
 ---@return integer? job_id
 local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, callback)
+  vim.api.nvim_exec_autocmds("User", {
+    pattern = "ConformPreFormatter",
+    data = { formatter = formatter },
```
stevearc, line 303: The whole formatter contains a lot of fields that probably aren't very useful. Could we start with just passing the name and then add other data if it's needed?

stevearc (changes requested):
I'm on board with the general idea; left a bit of feedback

gegoune:
To follow naming patterns of Vim/nvim I suggest changing event names to `ConformFormatPre` and `ConformFormatPost`.

See `:h autocmd-events`.

lua/conform/runner.lua
itmecho: Good point! I've left the structure of `data.formatter.name` so it can be easily extended in the future

lua/conform/runner.lua
itmecho: I've moved this into the callback and changed the event data to return the error (code and message)

itmecho:
> I'm on board with the general idea; left a bit of feedback

Thanks for the feedback! I have actioned that and also gone with the rename suggested by @gegoune

stevearc:
LGTM thanks for the PR!

PR #624, opened 2025-01-18
feat: added js-beautify (w/ css- & html-beautify)

magnuslarsen:
Added [js-beautify](https://github.com/beautifier/js-beautify) to conform.nvim, using [stdin](https://github.com/beautifier/js-beautify#options)

I took inspiration from the prettier config; though I haven't actually tested the "fs.is_windows and js-beautify.cmd" part (I don't have a Windows to test on)

js-beautify comes with two extra bins: `html-beautify` and `css-beautify`; which is just a [an easy interface](https://github.com/beautifier/js-beautify?tab=readme-ov-file#css--html) to `js-beautify` (passing in `--html` and `--css` respectively)

What do you think is the best way to implement this? They all take the same basic arguments, as such the 3 different implementations I made currently are all copy/paste, except for the binary name

Let me know what you think :-)


stevearc (approved):
I think this approach is probably best. Thanks for the PR!

PR #625, opened 2025-01-18
feat: added sqruff

magnuslarsen:
[sqruff](https://github.com/quarylabs/sqruff) is a promising new tool, that combines the speed of `ruff` with usefulness of `sqlfluff`

This implementation is heavily inspired by the [sqlfluff](https://github.com/stevearc/conform.nvim/blob/master/lua/conform/formatters/sqlfluff.lua) implementation (they are pretty much identical)

While sqruff works without a `.sqruff` file, it hardly makes sense to do so. So I've kept the enforcement of cwd


stevearc:
LGTM thanks!

PR #627, opened 2025-01-22
feat: support tables as keys in setup({ formatters_by_ft })

drowning-cat:
```lua
require('conform').setup({
  -- You can use tables as keys to define multiple file types for the same formatter config
  [{ "bash", "csh", "ksh", "sh", "zsh" }] = { "beautysh" },
  -- Simple keys always override tables as keys, regardless of order
  bash = { "shfmt" },
})
```
There is only one caveat: [lua-language-server](https://github.com/LuaLS/lua-language-server) currently does not support type annotations for table keys.
See https://github.com/LuaLS/lua-language-server/issues/2610 for more details.


drowning-cat:
I can also prioritize simple keys over tables-as-keys, if you'd like.

stevearc:
I appreciate the work that went into this, but I'm not interested in adding this syntax to `setup`. I have intentionally tried to make the interface as simple and stupid as possible, I've actually [_removed_ some old features](https://github.com/stevearc/conform.nvim/commit/9f111be14818c91832db8f320c4a4aa68de0e00b) because they complicated the API. It's easier to learn, easier to read, and is a hard guard against scope creep. I've had issues with my other plugins where feature request upon feature request causes the config options to explode. My current stance is that if it's something that can reasonably be done in the end-user config, it should be. Given that this is just a small bit of syntactic sugar to save a few lines, I don't think it's worth adding.

tanishqmanuja:
I am using this, would this cause any issues .. asking because what's in the commit utils looks different

```lua
function flatten_formatters(formatter_table)
    local flattened = {}
    
    for key, value in pairs(formatter_table) do
        if type(key) == "table" then
            -- Handle multiple (table)
            for _, filetype in ipairs(key) do
                flattened[filetype] = value
            end
        else
            -- Handle single (string)
            flattened[key] = value
        end
    end
    
    return flattened
end
```

drowning-cat:
Don't worry, there will be no problem at all, except in very rare cases where you *mutate* `formatters_by_ft` directly, because they will link to the same object. For example:

```lua
table.insert(require('conform').formatters_by_ft[<ft>], <additional-formatter>)
```

If you want to prevent this, simply make a deep copy of the value:
```lua
for _, filetype in ipairs(key) do
  flattened[filetype] = vim.deepcopy(value)
end
```

tanishqmanuja:
@drowning-cat, 

Thanks for the detailed explanation and clarity on table references, really appreciate it!

I am a js dev so still drawing parallels between it and lua 😅

tanishqmanuja:
@drowning-cat, 

Thanks for the detailed explanation and clarity on table references, really appreciate it!

I am a js dev so still drawing parallels between it and lua 😅

PR #629, opened 2025-01-23
feat: added pyproject-fmt

mehalter:
This adds support for [pyproject-fmt](https://github.com/tox-dev/toml-fmt/tree/main/pyproject-fmt) which is used to format `pyproject.toml` files in a Python project


stevearc:
Thanks for the PR!

PR #631, opened 2025-01-23
fix: prefer subfiletype over parent filetype

### LINKED ISSUE #630: bug: Incorrect formatter picked in cases of nested filetypes
### Neovim version (nvim -v)

NVIM v0.10.3

### Operating system/version

Arch Linux

### Read debugging tips

- [x] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

### Log file

With the file `test.hcl` opened:

```
Log file: /home/mhalter3@icl.gtri.org/Documents/nvim_sandbox/.repro//state/nvim/conform.log

Formatters for this buffer:
hcl unavailable: Command not found

Other formatters:
packer_fmt ready (packer) /usr/sbin/packer
```

With the file `test.pkr.hcl` opened:

```
Log file: /home/mhalter3@icl.gtri.org/Documents/nvim_sandbox/.repro//state/nvim/conform.log

Formatters for this buffer:
hcl unavailable: Command not found

Other formatters:
packer_fmt ready (packer) /usr/sbin/packer
```

### Describe the bug

Conform.nvim has support for detecting filetypes both root filetypes and nested filetypes such as `hcl` vs `hcl.packer`. So if I only configure `packer` in the `formatters_by_ft` then it correctly identifies `hcl.packer` as a compatible `packer` filetype. But I noticed if I configure both `hcl` and `hcl.packer` then the `hcl` formatters are the only ones that get attached to the buffer. I think the priority here should favor the more specific filetype.

breaking (some functionality is broken)

### Steps To Reproduce

1. `nvim -u repro.lua test.hcl`, Open up an `.hcl` file
2. `:ConformInfo` and see that `hcl` is set up
3. `:e test.pkr.hcl`, open up a `.pkr.hcl` file
4. `:ConformInfo` and see that `hcl` formatters are still used and the `packer` formatters are ignored.

_Note:_ If the `hcl` formatters are removed and `packer` only exists, then opening a `test.pkr.hcl` file and running `:ConformInfo` shows that the formatters are correctly picked up for `packer`.

### Expected Behavior

The formatter configured for the most specific filetype should be used. In this case, use the `packer` formatters for the filetype `hcl.packer` and the `hcl` formatters for the filetype `hcl` or any other `hcl.X` unless `X` is configured.

### Minimal example file

Just needs files `test.hcl` and `test.hcl.packer`. Contents doesn't matter.

Also the main reason for using this nested filetype notation is to get stuff like treesitter and LSP to work nicely with the files without having to manually tell each that those nested filetypes should also be treated like their parent filetype.

### Minimal init.lua

```Lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system "curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua")()

vim.keymap.set("n", "<c-a>", function()
  vim.ui.input({ prompt = "Enter text: " }, function(input) vim.notify(input) end)
end)

-- Add a new filetype for `.pkr.hcl` files
vim.filetype.add {
  pattern = {
    [".*%.pkr%.hcl"] = "hcl.packer",
  },
}

require("lazy.minit").repro {
  spec = {
    {
      "stevearc/conform.nvim",
      opts = {
        -- add your config here
        formatters_by_ft = {
          hcl = { "hcl" },
          packer = { "packer_fmt" },
        },
      },
    },
  },
}
```

### Additional context

In the example my logs show `hcl` command is not found. I just wanted to point out that this doesn't matter in the testing case just because I'm looking at how the buffer is configured and not having any problems with the formatters actually running.

Thanks so much for your maintenance of this plugin as it truly is an amazing formatting setup!

mehalter:
This changes the filetype iteration to prefer sub-filetypes over the parent filetypes.

Closes #630

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -239,7 +239,8 @@ local function get_matching_filetype(bufnr)
   end
   local filetypes = vim.split(vim.bo[bufnr].filetype, ".", { plain = true })
   table.insert(filetypes, "_")
-  for _, filetype in ipairs(filetypes) do
+  for idx = #filetypes, 1, -1 do
+    local filetype = filetypes[idx]
     local ft_formatters = M.formatters_by_ft[filetype]
     -- Sometimes people put an empty table here, and that should not count as configuring formatters
     -- for a filetype.


stevearc:
Oh lol sorry I didn't see this PR until I'd already put up 783110741669a52b85cfac41f499c8162c4423fa
Nice validation that it's the correct fix, though!

PR #639, opened 2025-02-04
feat: ansible-lint

G4bbix:
I set async since ansible lint takes a lot of time to execute, depending on the file/environment.

### DIFF
--- a/lua/conform/formatters/ansible-lint.lua
+++ b/lua/conform/formatters/ansible-lint.lua
@@ -6,10 +6,6 @@ return {
   },
   command = "ansible-lint",
   args = { "-f", "codeclimate", "-q", "--fix=all", "$FILENAME" },
-  options = {
-    ignore_errors = true,
-    async = true,
-  },
   stdin = false,
-  exit_codes = { 0, 1, 2, 3, 4 },
+  exit_codes = { 0, 2 },
 }
--- a/lua/conform/formatters/ansible-lint.lua
+++ b/lua/conform/formatters/ansible-lint.lua
@@ -1,0 +1,11 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/ansible/ansible-lint",
+    description = "ansible-lint with --fix",
+  },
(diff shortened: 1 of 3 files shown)


lua/conform/formatters/ansible-lint.lua
```diff
@@ -0,0 +1,15 @@
...
+    description = "ansible-lint with --fix",
+  },
+  command = "ansible-lint",
+  args = { "-f", "codeclimate", "-q", "--fix=all", "$FILENAME" },
+  options = {
+    ignore_errors = true,
+    async = true,
+  },
+  stdin = false,
+  exit_codes = { 0, 1, 2, 3, 4 },
```
stevearc, line 14: I found this reference for the exit codes: https://github.com/ansible/ansible-lint/blob/9391e676532b92fdbfd02a3dc9cf5ff3d50daffc/src/ansiblelint/constants.py#L36-L40
That doesn't seem to match up with what you have here.

lua/conform/formatters/ansible-lint.lua
```diff
@@ -0,0 +1,15 @@
...
+  meta = {
+    url = "https://github.com/ansible/ansible-lint",
+    description = "ansible-lint with --fix",
+  },
+  command = "ansible-lint",
+  args = { "-f", "codeclimate", "-q", "--fix=all", "$FILENAME" },
+  options = {
+    ignore_errors = true,
+    async = true,
+  },
```
stevearc, line 12: I don't know what you think this is doing, but some formatters specify options that are then used in the `command` or `args` callbacks. Nothing here reads the options, so they will have no effect.

stevearc (changes requested):

lua/conform/formatters/ansible-lint.lua
G4bbix: You are right, I had no idea what I was doing. I removed them

lua/conform/formatters/ansible-lint.lua
G4bbix: Thanks, I adjusted the return codes according to the information provided

stevearc:
LGTM thanks!

Issue #275, opened 2024-01-19
bug: styler fails


6801318d8d:
### Neovim version (nvim -v)

0.9.5

### Operating system/version

Arch Linux

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

10:57:10[DEBUG] Running formatters on test.R: { "styler" }
10:57:10[INFO] Run styler on test.R
10:57:10[DEBUG] Creating temp file .conform.7244551.test.R
10:57:10[DEBUG] Run command: { "R", "-s", "-e", "r.nvim::format()", "--args", ".conform.7244551.test.R" }
10:57:10[DEBUG] styler exited with code 0
10:57:10[DEBUG] Cleaning up temp file .conform.7244551.test.R

### Describe the bug

1. Formatting R code has no effect

2. r.nvim is working, and the command line that conform.nvim is trying to run is working as well:
```
➜ # see minimal example file for test.R
➜ R -s -e "r.nvim::format()" --args ./test.R
df <- function(a, b) {
	print("ciao")
	c <- a + b
	print(paste0(c))
}
```
3. Different from #242, in which the error message says that `r.nvim` is not installed

### What is the severity of this bug?

breaking (some functionality is broken)

### Steps To Reproduce

See above

### Expected Behavior

R code should be formatted, code inside function should be indented

### Minimal example file

df <- function(a,b) {
print("ciao")
c = a+b
print(paste0(c))
}

### Minimal init.lua

```Lua
-- conform.nvim
require("conform").setup({
	-- Set the log level. Use `:ConformInfo` to see the location of the log file.
	log_level = vim.log.levels.DEBUG,
	formatters = {
		isort = {
			command = "/usr/bin/isort",
		},
		black = {
			command = "/usr/bin/black",
		},
	},
	formatters_by_ft = {
		python = { "isort", "black" },
		lua = { "stylua" },
		-- styler requires installing the R package `r.nvim` in R
		-- Open R and execute `remotes::install_github("devOpifex/r.nvim")`
		r = { "styler" },
	},
})

vim.api.nvim_create_user_command("ConformFormat", function(args)
	local range = nil
	if args.count ~= -1 then
		local end_line = vim.api.nvim_buf_get_lines(0, args.line2 - 1, args.line2, true)[1]
		range = {
			start = { args.line1, 0 },
			["end"] = { args.line2, end_line:len() },
		}
	end
[...]

PR #641, opened 2025-02-05
use styler natively for R instead of calling it via r.nvim::format

kikur4ge:
Currrently the styler command in conform relies on the user using R.nvim and having the associated r package installed. However, under the hood, r.nvim just calls styler directly. I have updated the default styler configuration to do the same. I have also added some comments describing usage. This is related to #275.


stevearc (commented):
Not being familiar with the R ecosystem myself, I can't really validate if these changes work/are reasonable. I trust that the R formatting community will file issues if there's anything wrong here.

PR #642, opened 2025-02-06
feat: add codeql native formatter

snowdroppe:
Added support for the codeql query formatter. Documentation can be found [here](https://docs.github.com/en/code-security/codeql-cli/codeql-cli-manual/query-format)


stevearc (approved):
Thanks for the PR!

PR #657, opened 2025-02-18
feat: add runic formatter

fredrikekre:
This adds support for [Runic.jl](https://github.com/fredrikekre/Runic.jl) -- a code formatter for Julia source code.


stevearc:
Thanks for the PR!

PR #658, opened 2025-02-19
fix: fix uncaught undojoin after undo error

sid-6581:
Catch the "undojoin after undo" error the same way as the other call to `vim.cmd.undojoin`.


stevearc:
LGTM thanks!

PR #662, opened 2025-02-23
Enable cabal-fmt to process expand pragmas

lrworth:
A comment like

```cabal
-- cabal-fmt: expand src/
```

cannot be processed without a file path, which means the filename needs
to be passed as input.


stevearc:
LGTM thanks!

PR #663, opened 2025-02-28
fix(prettypst): stdin and stdout args

FlyingStitchman:
This tells prettypst to use stdin and stdout for the formatting. Before it errored out and did not format the buffer.


stevearc:
Sure, LGTM!

Issue #541, opened 2024-10-08
bug: `csharpier` does not work anymore


KennethHoff:
### Neovim version (nvim -v)

NVIM v0.10.2

### Operating system/version

Arch Linux

### Read debugging tips

- [X] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

Not relevant

### Describe the bug

`dotnet-charpier` is no longer a valid command. You have to use `dotnet` with the argument `sharpier` (and then pass whatever argument afterwards.

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

Call `:ConformInfo`

### Expected Behavior

		formatters = {
			csharpier = {
				command = "dotnet",
				args = { "csharpier", "--write-stdout", "$FILENAME" },
				stdin = true,
			},
		},

### Minimal init.lua

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/conform.nvim",
    config = function()
      require("conform").setup({
        log_level = vim.log.levels.DEBUG,
        -- add your config here
      })
    end,
opts = {
		formatters_by_ft = {
			cs = { "csharpier" },
		},
		format_on_save = {
			timeout_ms = 5000,
		},
	},
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Additional context

Updated in this PR: https://github.com/belav/csharpier/pull/1332

PR #668, opened 2025-03-12
update csharpier fomatter for speed

fanlia:
use dotnet-csharpier instead


stevearc:
I no longer write C#, so one of you C# devs is going to have to tell me who is correct here. According to #541 dotnet-csharpier is no longer a valid command. Why is that?

fanlia:
maybe I am using global tool
```sh
dotnet tool list -g csharpier

Package Id      Version      Commands        
---------------------------------------------
csharpier       0.30.6       dotnet-csharpier
```

PR #671, opened 2025-03-14
Added support of toolchains for rustfmt

mikebohdan:
I added the ability to pick a toolchain for rustfmt. This will allow nightly formatting rules while using a non-nightly toolchain.

### DIFF
--- a/lua/conform/formatters/rustfmt.lua
+++ b/lua/conform/formatters/rustfmt.lua
@@ -10,9 +10,11 @@ return {
   options = {
     -- The default edition of Rust to use when no Cargo.toml file is found
     default_edition = "2021",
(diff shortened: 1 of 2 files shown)


lua/conform/formatters/rustfmt.lua
```diff
@@ -10,9 +10,11 @@ return {
   options = {
     -- The default edition of Rust to use when no Cargo.toml file is found
     default_edition = "2021",
+    default_toolchain = "stable",
```
stevearc, line 13: The `default_edition` is only a default because it's what we use if there is no Cargo.toml file. Let's rename this to just `toolchain`.

stevearc (changes requested):

PR #674, opened 2025-03-21
Fix syntax_tree for ignored files

seven1m:
First, thank you for this project.

I wanted to suggest this change for the `syntax_tree` formatter based on a project I'm working on.

Running the command with the actual relative path (not a temp path) allows it to honor any `--ignore-files` configuration in the .streerc file. It is true that we're actually ignoring the stdin, so that might make this change a non-starter.

But it does work. Given an `.streerc` config like this:

```
--ignore-files=bar.rb
```

The `bar.rb` file will not be formatted.


seven1m:
I withdraw my change. It doesn't work for unsaved changes in the file. :sob:

PR #678, opened 2025-03-27
feat: add golangci-lint fmt

dro-sh:
Add new formatter: golangci-lint fmt. It was released 3 days ago.


kndndrj:
Can confirm this works for me.

Also note that the user will have to have their own `.golangci.yml` or use `append_args` option to enable formatters, since they are all disabled by default.

dro-sh:
@kndndrj `golangci-lint` is de-facto standard tool for linting, and most of projects have `.golangci.yaml`. But there are cases then projects could not have config, e.g. local small testing project. `golangci-lint fmt` has option `--config`, but I think this option overrides config from root of project. It is nice if `golangci-lint fmt` will have option `--fallback-config` which would use config from path which are set by user if config isn't found at root of project

stevearc:
Thanks for the PR!

PR #683, opened 2025-04-02
fix(sqruff): update args, stdin and require_cwd

whisperpine:
The arg "--force" must be used for auto formatting, otherwise it logs error in ConformInfo when trying to format.

Reading from stdin seams not working,
thus stdin is set to false and "$FILENAME" is used instead.

sqruff doesn't require cwd (the ".sqruff" file).
Without ".sqruff", it uses the default config.


stevearc:
LGTM thanks!

PR #684, opened 2025-04-02
feat: add support for mago

hougesen:
Adds support for the `mago format`  and `mago lint` commands.


stevearc (changes requested):
nit: could we name these `mago_format` and `mago_lint` instead? It makes it slightly easier to override the settings in config because you don't have to wrap it in a table key like `["mago-format"]`

hougesen:
Sure, makes sense 😄

stevearc:
Great, thanks for the PR!

PR #686, opened 2025-04-05
add formatter definition for odinfmt

BlameTroi:
A definition for the odinfmt program in formatters.


stevearc:
Thanks for the PR!

PR #691, opened 2025-04-18
fix: respect dos line endings

Gatsik:
I encountered a very annoying issue with `isort` on Windows -- formatter always added unnecessary line after imports.

For example:
```diff
-import sys
 import os
+import sys
+
```
or
```diff
-import sys
 import os
+import sys


 def main() -> int:
+
     return 0
```
(added empty lines are redundant)

It looks like runner doesn't properly handle final `\r` on processing stdout:
```log
[TRACE] Applying text edits: { {
              newText = "os\r\nimport sys\r\n\r\n\r\ndef main() -> int:\r",
              range = {
               ...
              }
            } }
```

I came up with 2 solutions: the first one is this PR and the second one is simple:
```diff --git a/lua/conform/runner.lua b/lua/conform/runner.lua
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -153,6 +153,7 @@ local function create_text_edit(
     table.insert(replacement, "")
   end
-  local new_text = table.concat(replacement, "\n")
+  local new_text = table.concat(replacement, "\n"):gsub("\r$", "")
 
   return {
     newText = new_text,
```

I don't know which one is fine (maybe neither!), but the more complex one ended up as a PR because it edits multiple files


stevearc:
This looks good to me. Thanks for the fix!

PR #698, opened 2025-05-01
fix(csharpier): support new cli for csharpier 1.0.x

jmkelly:
with csharpier going to version 1 it has changed the api.  This will be a breaking change if people aren't running on csharpier 1.0.x.


jmkelly:
didn't check that there was already a PR for this...and it was better so I've closed mine

Issue #701, opened 2025-05-07
feature request: add treefmt formatter


DockterTeagle:
### Did you check existing requests?

- [x] I have searched the existing issues

### Describe the feature

add the treefmt formatter

### Provide background

this is a little bit of a weird request as treefmt is meant to.. well format the entire tree, but it can work with single files, and this allows you to use one formatter for all filetypes, instead of installing a formatter twice(kindof). If you dont know treefmt is an "all in one" formatter that implements other formatters.

### What is the significance of this feature?

nice to have

### Additional details

I have made a custom formatter for now,, and can submit a PR after I read the codebase more

hougesen:
This was closed by #702

DockterTeagle:
sorry forgot to close this issue I opened, that I then resolved

PR #702, opened 2025-05-07
feat: inital support for treefmt formatter

DockterTeagle:
https://github.com/stevearc/conform.nvim/issues/701
idk if this is wanted at all, but treefmt does have its own option I need to work on, but treefmt is designed to be per-repo?


stevearc:
Thanks for the PR!

PR #707, opened 2025-05-10
formatters: add dioxus (`dx fmt`)

kasimeka:
i prefer to use it with `require_cwd = true` and the config snippet below, but i think they're too much of a personal preference to include in the pr

```lua
formatters_by_ft.rust = function()
  if vim.fn.executable("dx") == 1 then
    return { "rustfmt", "dioxus" }
  end
  return { "rustfmt" }
end
```


stevearc:
Thanks for the PR!

PR #711, opened 2025-05-21
Add .cljstyle as the root file for cljstyle formatter

sonntag:
`cljstyle` uses a hierarchy of `.cljstyle` configuration files in the source tree to change formatting behavior, so if it is run in the wrong directory the formatting may be incorrect.

This sets the cwd to one that contains a `.cljstyle` file.

https://github.com/greglook/cljstyle?tab=readme-ov-file#configuration


stevearc:
LGTM thanks!

PR #721, opened 2025-06-04
Update csharpier after breaking change in 1.x release

thomaslaich:
As of release 1, csharpier slightly reworked their CLI syntax (see https://github.com/belav/csharpier/releases/tag/1.0.0).

We now have to type:
```bash
dotnet csharpier format <file>
```
instead of
```bash
dotnet csharpier <file>
```

Thus, we have to adjust the command in conform.


thomaslaich:
ups, i made a duplicate:

https://github.com/stevearc/conform.nvim/pull/695

closing this in favor of the other more complete PR.

PR #723, opened 2025-06-06
feat(recipes): organize imports using LSP before format

shljsl75891:
- recipe to organize imports using `ts_ls` LSP before formatting

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -107,6 +107,54 @@ end, {
 })
 ```
 
+## Organize Imports usign `ts_ls` language server before format
+
+Execute LSP commands before formatting file. Timeout can be reduced if using `prettierd` formatter.
+
+```lua
+require("conform").setup({
+    format_on_save = function(bufnr)
+        if vim.g.disable_autoformat or vim.b[bufnr].disable_autoformat then
+            return
+        end
+        local TIMEOUT = 2000
+        local bufname = vim.api.nvim_buf_get_name(bufnr)
+        if bufname:match("/node_modules/") then
+            return
+        end
+        local opts = {
+            quiet = true,
+            timeout_ms = TIMEOUT,
+            lsp_format = "fallback",
+        }
+        if
+            vim.tbl_contains({
+                "javascript",
+                "javascriptreact",
+                "typescript",
+                "typescriptreact",
+            }, vim.bo[bufnr].filetype)
+        then
+            local client = vim.lsp.get_clients({ name = "ts_ls" })[1]
+            if client then
+                local pos = vim.api.nvim_win_get_cursor(0)
+                client:exec_cmd({
+                    title = "",
+                    command = "_typescript.organizeImports",
+                    arguments = { vim.api.nvim_buf_get_name(0) },
+                }, { bufnr = 0 }, function()
+                        require("conform").format(opts)
+                        vim.cmd("noautocmd update")
+                        vim.api.nvim_win_set_cursor(0, pos)
+                    end)
+            end
+        else
+            return opts
+        end
+    end,
+})
+```
+
 ## Lazy loading with lazy.nvim
 
 Here is the recommended config for lazy-loading using lazy.nvim


doc/recipes.md
```diff
@@ -107,6 +108,61 @@ end, {
...
+            javascriptreact = true,
+            typescript = true,
+            typescriptreact = true,
+        }
+
+        if ts_filetypes[filetype] then
+            local client = vim.lsp.get_clients({ name = "ts_ls" })[1]
+            if client then
+                local prev_pos = vim.api.nvim_win_get_cursor(0)
+                client:exec_cmd({
```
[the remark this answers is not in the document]
shljsl75891: when I was fixing this, I saved the file with `vim.cmd("noautocmd update")`. Now, used different approach. Please re-review

doc/recipes.md
```diff
@@ -107,6 +108,61 @@ end, {
...
+    format_on_save = function(bufnr)
+        if vim.g.disable_autoformat or vim.b[bufnr].disable_autoformat then
+            return
+        end
+
+        local bufname = vim.api.nvim_buf_get_name(bufnr)
+        if bufname:find("/node_modules/", 1, true) then
+            return
+        end
+
```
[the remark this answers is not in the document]
shljsl75891: Yes, I agree. Now, kept only the relevant part to the recipe. Please re-review...

doc/recipes.md
```diff
@@ -107,6 +108,61 @@ end, {
...
+            javascript = true,
+            javascriptreact = true,
+            typescript = true,
+            typescriptreact = true,
+        }
+
+        if ts_filetypes[filetype] then
+            local client = vim.lsp.get_clients({ name = "ts_ls" })[1]
+            if client then
+                local prev_pos = vim.api.nvim_win_get_cursor(0)
```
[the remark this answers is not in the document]
shljsl75891: The idea behind this was the `organize imports` command repaints the buffer, and move the cursor automatically at the end of imports. It was annoying, so I want my cursor to remain close to previous position. I changed the code, and tested `:wall` command, the imports are being organized fine and cursor is restoring its previous position. Please re-review

doc/recipes.md
```diff
@@ -107,6 +108,61 @@ end, {
...
+        end
+
+        local bufname = vim.api.nvim_buf_get_name(bufnr)
+        if bufname:find("/node_modules/", 1, true) then
+            return
+        end
+
+        local filetype = vim.bo[bufnr].filetype
+        local opts = {
+            quiet = true,
```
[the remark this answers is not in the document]
shljsl75891: ok sure, removed from the example.

doc/recipes.md
```diff
@@ -107,6 +108,43 @@ end, {
...
+You can run LSP commands before formatting a file. This example uses the `ts_ls` server to organize imports along with formatting, but the structure is adaptable to any LSP action.
+
+```lua
+vim.api.nvim_create_autocmd("BufWritePre", {
+	desc = "Format before save",
+	pattern = "*",
+	group = vim.api.nvim_create_augroup("FormatConfig", { clear = true }),
+	callback = function(ev)
+        local conform_opts = { bufnr = ev.buf, lsp_format = "fallback", timeout_ms = 2000 }
+        local bufname = vim.api.nvim_buf_get_name(ev.buf)
```
stevearc, line 122: unused?

doc/recipes.md
```diff
@@ -107,6 +108,43 @@ end, {
...
+        local client = vim.lsp.get_clients({ name = "ts_ls", bufnr = ev.buf })[1]
+        local ts_filetypes = { javascript = true, javascriptreact = true, typescript = true, typescriptreact = true }
+
+        if not ts_filetypes[vim.bo[ev.buf].filetype] or not client then
```
stevearc, line 126: Probably don't need the filetype checks here, right? As long as `ts_ls` is attached we should be able to organize the imports

doc/recipes.md
```diff
@@ -107,6 +108,43 @@ end, {
...
+            return
+        end
+
+        client:request("workspace/executeCommand", {
+            command = "_typescript.organizeImports",
+            arguments = { vim.api.nvim_buf_get_name(ev.buf) },
+        }, function(err)
+                if err then
+                    -- error handling if command fails
+                    return
```
stevearc, line 137: Let's add a `vim.notify` here. It may be noisy in practice, but it's a good default for users to get _some_ feedback from a failure and they can always tweak it themselves if they dislike it.

doc/recipes.md
```diff
@@ -107,6 +108,43 @@ end, {
...
+        local bufname = vim.api.nvim_buf_get_name(ev.buf)
+        local client = vim.lsp.get_clients({ name = "ts_ls", bufnr = ev.buf })[1]
+        local ts_filetypes = { javascript = true, javascriptreact = true, typescript = true, typescriptreact = true }
+
+        if not ts_filetypes[vim.bo[ev.buf].filetype] or not client then
+            require("conform").format(conform_opts)
+            return
+        end
+
+        client:request("workspace/executeCommand", {
```
stevearc, line 131: I thought you said that you replaced this with `request_sync`? I think this will have the same issue I mentioned before

doc/recipes.md
```diff
@@ -107,6 +108,43 @@ end, {
...
+            arguments = { vim.api.nvim_buf_get_name(ev.buf) },
+        }, function(err)
+                if err then
+                    -- error handling if command fails
+                    return
+                end
+                require("conform").format(conform_opts)
+                if vim.bo.buftype == "" then
+                    vim.cmd("silent noautocmd write")
+                end
```
stevearc, line 142: This shouldn't be necessary if you use `client:request_sync`

stevearc (changes requested):

doc/recipes.md
shljsl75891: `client:request_sync` this is synchronous but, and blocks write until imports are organized

doc/recipes.md
shljsl75891: `request_sync` is blocking and synchronous. So, if LSP is initializing, then save won;t happen until LSP is fully initialized. But, I am only formatting in callback of this method

doc/recipes.md
shljsl75891: removed

doc/recipes.md
shljsl75891: added.

doc/recipes.md
shljsl75891: yes correct. only `client` check would be enough. Fixed

doc/recipes.md
shljsl75891: If you want, I can implement that as follows
```lua
local client = vim.lsp.get_clients({ name = "ts_ls", bufnr = ev.buf })[1]

if not client then
	require("conform").format(conform_opts)
	return
end

local request_result = client:request_sync("workspace/executeCommand", {
	command = "_typescript.organizeImports",
	arguments = { vim.api.nvim_buf_get_name(ev.buf) },
}, timeout)

if request_result and request_result.err then
	vim.notify(request_result.err.message, vim.log.levels.ERROR)
	return
else
	require("conform").format(conform_opts)
end
```

doc/recipes.md
shljsl75891: This was an intentional change — I used client:request because client:request_sync blocks the write until imports are organized. It was annoying for me when experimenting in simple files, but since this is the desired behavior for the plugin, I kept it synchronous here. Any user can modify the recipe according to his choice.

stevearc:
This looks good! Thanks for the contribution!

PR #737, opened 2025-06-25
feat: add pyupgrade

filipgodlewski:
feat: add pyupgrade

### DIFF
--- a/lua/conform/formatters/pyupgrade.lua
+++ b/lua/conform/formatters/pyupgrade.lua
@@ -6,7 +6,7 @@
 return {
   meta = {
     url = "https://github.com/asottile/pyupgrade",
-    description = "A tool (and pre-commit hook) to automatically upgrade syntax for newer versions of the language.",
+    description = "A tool to automatically upgrade syntax for newer versions of Python.",
   },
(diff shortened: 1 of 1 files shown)


filipgodlewski:
@stevearc it would actually really be nice to add to the `meta` some sort of `notes` or `implementation remarks` so that people get it up front when they read the docs about a particular formatter. What do you think?

filipgodlewski:
> Thanks for the PR!
> 
> 
> 
> If you want you could add an option for this [like we do for rustfmt](https://github.com/stevearc/conform.nvim/blob/b529dd4897c85c3188cc787084089a9d55843093/lua/conform/formatters/rustfmt.lua#L10-L13), then it will show up [in this documentation](https://github.com/stevearc/conform.nvim/blob/master/doc/formatter_options.md). I'm not too concerned about it, though. My philosophy is if you're using a formatter, you should know what it is and how it works.

Yeah, I thought about it, but then again I don't want to force anyone for a particular version. Pyugrade falls back to `--py3` anyways + I think that I saw discussions about using the current environment python version, so yeah, let's maybe leave it as it is 😏

PR #740, opened 2025-06-29
feat: add tailwindcss-language-formatter

xNyfPtx:
i have no idea what to add inside `args` . You can just check the lsp on `meta.url`. Also, are there any docs for contributors about adding formatters?

### DIFF
--- a/lua/conform/formatters/tailwindcss-language-server.lua
+++ b/lua/conform/formatters/tailwindcss-language-server.lua
@@ -1,0 +1,9 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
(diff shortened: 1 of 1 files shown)


stevearc:
This looks like a language server, not a formatter. Conform is for use with standalone formatting commands like `rustfmt` and `black`, not LSPs like `rust-analyzer` and `pyright`. Neovim already has built-in LSP integration, including formatting. If you are connected to the tailwind LSP, `vim.lsp.buf.format()` will do the formatting. You do also have the option to trigger LSP formatting _through_ conform, which has a couple advantages that are documented in the README

Issue #192, opened 2023-11-10
feat: Command to disable formatting


BitInByte:
Hello @stevearc and all maintainers, thanks for the amazing work, not only on conform.nvim but also on oil.nvim which I heavily use them. 

Would it be possible to add a command to conform to disable the format on save? I used that a lot with null-ls, sometimes my formatting rules are different than the repo rules and sometimes I get a diff on git logs just for formatting changes and what I was doing in this cases on null-ls were disabling the formatting, save the file and then enabling the formatting again.

Would it be possible to do with conform as well?

Thanks in advance :D

stevearc:
I don't think this makes sense to add to the core plugin, but it's easy enough to add this logic yourself! https://github.com/stevearc/conform.nvim/blob/master/doc/recipes.md#command-to-toggle-format-on-save

andradei:
Wanting to write a buffer without formatting is a valid use-case. Specially when you're working on a code base in a team and you don't want your formatter to change the entire file because a line change you had to make.

breakingram:
Here is my auto-format toggle config:

`<leader>tf` - will toggle per‑buffer autoformat
`<leader>tF` - will toggle global autoformat

```lua
return { -- Autoformat
  'stevearc/conform.nvim',
  lazy = false,
  keys = {
    {
      '<leader>tf',
      function()
        -- If autoformat is currently disabled for this buffer,
        -- then enable it, otherwise disable it
        if vim.b.disable_autoformat then
          vim.cmd 'FormatEnable'
          vim.notify 'Enabled autoformat for current buffer'
        else
          vim.cmd 'FormatDisable!'
          vim.notify 'Disabled autoformat for current buffer'
        end
      end,
      desc = 'Toggle autoformat for current buffer',
    },
    {
      '<leader>tF',
      function()
        -- If autoformat is currently disabled globally,
        -- then enable it globally, otherwise disable it globally
        if vim.g.disable_autoformat then
          vim.cmd 'FormatEnable'
          vim.notify 'Enabled autoformat globally'
        else
          vim.cmd 'FormatDisable'
          vim.notify 'Disabled autoformat globally'
        end
      end,
      desc = 'Toggle autoformat globally',
    },
  },
  opts = {
    notify_on_error = false,
    format_on_save = function(bufnr)
      if vim.g.disable_autoformat or vim.b[bufnr].disable_autoformat then
        return
      end
      local disable_filetypes = { c = false, cpp = false }
      return {
[...]

PR #742, opened 2025-06-30
(feat): add commands to toggle conform's format on save

ingenarel:
i use autoformat on autosave in my personal neovim config with custom autocommands (i do some regex stuff after post format), but sometimes stuff breaks, and then i need to manually disable the autosave function from my config, but i already fixed it in my personal config by making some use commands to turn it off and on. So i thought why not just implement it on the main repo too

i added two handy little commands, `:ConformEnable` and `:ConformDisable` which only gets set if you use the format-on-save functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,9 @@ require("conform").setup({
 })
 ```
 
+If you use this, conform will also set up two commands so you can turn on and off the format-on-save functionality.
+`:ConformEnable` and `:ConformDisable`
+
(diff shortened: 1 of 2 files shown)


stevearc:
I appreciate the PR, but this is not functionality that I'm interested in adding to conform. These are provided as recipes because they are best left to the end user's config. The reasoning is explained more in depth here: https://github.com/stevearc/conform.nvim/issues/192#issuecomment-2817321950

PR #746, opened 2025-07-06
fix(biome): respect buffer's expandtab/shiftwidth

phanen:


stevearc:
That's a nice adaptation, thanks for the PR!

PR #747, opened 2025-07-07
fix(rustfmt): workspace edition not found

FrancescElies:
Imagine you have a project part of a worksapce with the following definition `edition = { workspace = true }` in that case one needs to continue the search in the upper folders to find the workspace `Cargo.toml`.

This pr solves that.

### DIFF
--- a/lua/conform/util.lua
+++ b/lua/conform/util.lua
@@ -193,20 +193,15 @@ M.parse_rust_edition = function(dir)
   local manifest_files = vim.fs.find("Cargo.toml", { upward = true, path = dir, limit = math.huge })
   for _, manifest in ipairs(manifest_files) do
     for line in io.lines(manifest) do
-      -- if a project is part of a workspace edition might be defined top-level
-      local matches = line:match("edition *= *{ *workspace *= *true *}")
+      -- if a project is part of a workspace, the edition might be defined top-level
+      local is_in_workspace = line:match("edition *= *{ *workspace *= *true *}")
         or line:match("edition.workspace *= *true")
-      if matches then
-        goto upper_cargo_toml
+      local edition = line:match("^edition *= *(%d+)")
+      if is_in_workspace then
+        break
+      elseif edition then
+        return edition
       end
-      -- read cargo.toml edition
-      if line:match("^edition *=") then
-        local edition = line:match("%d+")
-        if edition then
-          return edition
-        end
-      end
-      ::upper_cargo_toml::
     end
   end
 end


lua/conform/util.lua
```diff
@@ -193,6 +193,13 @@ M.parse_rust_edition = function(dir)
   local manifest = vim.fs.find("Cargo.toml", { upward = true, path = dir })[1]
   if manifest then
     for line in io.lines(manifest) do
+      -- if a project is part of a workspace edition might be defined top-level
+      local matches = line:match("edition *= *{ *workspace *= *true *}")
+        or line:match("edition.workspace *= *true")
+      if matches then
+        local upper_folder = vim.fn.fnamemodify(manifest, ":h:h")
+        return M.parse_rust_edition(upper_folder)
```
stevearc, line 201: Instead of recursing, can you rewrite this so that `vim.fs.find` passes in `limit = math.huge` and iterates over all of the results?

lua/conform/util.lua
```diff
@@ -193,6 +193,13 @@ M.parse_rust_edition = function(dir)
   local manifest = vim.fs.find("Cargo.toml", { upward = true, path = dir })[1]
   if manifest then
     for line in io.lines(manifest) do
+      -- if a project is part of a workspace edition might be defined top-level
+      local matches = line:match("edition *= *{ *workspace *= *true *}")
+        or line:match("edition.workspace *= *true")
```
stevearc, line 198: A question that comes from not being familiar with cargo workspaces: do we really care if the value of `edition.workspace` is `true` or should we _always_ look for the value of the edition in parent `Cargo.toml` files if we fail to parse it from this file?

stevearc (changes requested):
1 req, 1 question

lua/conform/util.lua
FrancescElies: I think we do care, i believe values are not implicitly read from upper toml, but not 100% sure atm.


In the coming week I’ll search for the right parts of the docs or try it out myself and report back.

lua/conform/util.lua
FrancescElies: done

lua/conform/util.lua
FrancescElies: NOTE: I noticed that conform's default is 2021 which seems different than what rust chose.

stevearc:
Thanks for the PR!
I shuffled some things around a bit and removed the goto, since that's only available in LuaJIT and _technically_ there are some platforms that Neovim is built for that don't support it. If I bunged up any of the logic please either comment here or open a new PR

PR #751, opened 2025-07-15
fix: show errors from LSP responses instead of silently ignoring them

dmitmel:
The error returned by the `vim.lsp.Client:request_sync()` method only tells whether there was a timeout or an interrupt while waiting for the response with `vim.wait()`, the actual error returned by the LSP server is in `result.err`. This might make the LSP formatter slightly noisier than before though, because before it displayed errors only on timeouts, but hopefully that will not be much of a problem.


stevearc:
LGTM thanks!

utkarshgupta137:
This doesn't actually check if there is an error or not. It is breaking rust-analyzer.

PR #753, opened 2025-07-22
feat(formatters): add pymarkdownlnt

gsalvatella:
[pymarkdownlnt] is the python equivalent to the obiquitous markdownlint. For python projects, it makes more sense to use the former in order to avoid pulling Node.js and its dependencies. Add it to the official formatters.

[pymarkdownlnt]: https://github.com/jackdewinter/pymarkdown

### DIFF
--- a/lua/conform/formatters/pymarkdownlnt.lua
+++ b/lua/conform/formatters/pymarkdownlnt.lua
@@ -1,0 +1,10 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/jackdewinter/pymarkdown",
+    description = "PyMarkdown is primarily a Markdown linter",
+  },
+  command = "pymarkdownlnt",
+  args = { "--return-code-scheme", "minimal", "fix", "$FILENAME" },
+  stdin = false,
+}
--- a/lua/conform/formatters/pymarkdownlnt.lua
+++ b/lua/conform/formatters/pymarkdownlnt.lua
@@ -2,7 +2,7 @@
 return {
   meta = {
     url = "https://github.com/jackdewinter/pymarkdown",
-    description = "PyMarkdown is primarily a Markdown linter",
+    description = "A markdown linter and formatter.",
   },
   command = "pymarkdownlnt",
   args = { "--return-code-scheme", "minimal", "fix", "$FILENAME" },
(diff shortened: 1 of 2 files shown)


lua/conform/formatters/pymarkdownlnt.lua
```diff
@@ -0,0 +1,11 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/jackdewinter/pymarkdown",
+    description = "PyMarkdown is primarily a Markdown linter",
+  },
+  command = "pymarkdownlnt",
+  args = { "fix", "$FILENAME" },
+  exit_codes = { 0, 3 }, -- code 0 is given when trying a file that includes non-autofixable errors, code 3 when fixing is successfully applied
```
stevearc, line 9: What do you think of using [--return-code-scheme](https://pymarkdown.readthedocs.io/en/latest/user-guide/#-return-code-scheme-observability) instead of supplying special success exit codes?

stevearc (changes requested):

lua/conform/formatters/pymarkdownlnt.lua
gsalvatella: This is how the *default* scheme already behaves (returns 3 when it has fixed at least a file). Would you prefer to switch to the *minimal* scheme?

gsalvatella: Here you go :smile:

lua/conform/formatters/pymarkdownlnt.lua
```diff
@@ -0,0 +1,11 @@
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://github.com/jackdewinter/pymarkdown",
+    description = "PyMarkdown is primarily a Markdown linter",
+  },
+  command = "pymarkdownlnt",
+  args = { "--return-code-scheme", "minimal", "fix", "$FILENAME" },
+  exit_codes = { 0, 1 },
```
stevearc, line 9: Do we want this? According to the docs with `--return-code-scheme minimal` exit code 1 means `SYSTEM_ERROR`. I don't think that should be treated as a success.

stevearc (changes requested):

lua/conform/formatters/pymarkdownlnt.lua
gsalvatella: My apologies. I guess I did not understand that `exit_codes` refers only to the exit codes of successful execution, if this is what you're implying, but rather "possible exit codes you may encounter when running this command". I have removed the line entirely. Hope this is correct now?

stevearc:
LGTM thanks for the PR!

PR #757, opened 2025-07-27
feat: add qmlformat for QML files

peoxin:
Add [`qmlformat`](https://doc.qt.io/qt-6//qtqml-tooling-qmlformat.html) formatter for QML files.


stevearc:
Thanks for the PR!

PR #758, opened 2025-07-29
Don't overwrite given range with current selection

julienvincent:
Currently the `opts.range` argument is overwritten if the current mode is `V`.

This is due to missing logical grouping brackets on the `or` branch of the or check.


stevearc:
Thanks for the fix!

PR #759, opened 2025-08-02
feat(prolog): adding a prolog formatter for SWI-Prolog

kumkee:
Added prolog formatter from https://github.com/jamesnvc/lsp_server


stevearc:
Thanks for the PR!

PR #760, opened 2025-08-04
Update csharpier args to latest version

paul-manole:
Csharpier introduced a command `format` that is missing from current config which is causing the formatting to fail.


maxrzaw (approved):
I tested this out on csharpier 1.1.1 and it is working.

stevearc:
This was fixed in #695

PR #771, opened 2025-09-10
Added missing --stdin-filepath flag to Taplo formatter

mmottl:
This PR adds the missing `--stdin-filepath` flag to the Taplo formatter.

Taplo formatter configuration files (`.taplo.toml`, `taplo.toml`) may contain rules that are specific to file patterns. If Taplo does not receive the `--stdin-filepath` flag, it will not apply these rules.


stevearc:
Thanks for the fix!

PR #772, opened 2025-09-12
fix(csharpier): check if the dotnet command is available

Picalines:
`:ConformInfo` failed for me because i have `csharpier` installed via `mason.nvim`, but have no `dotnet` on my working machine

This commit adds a check for this case while keeping the local `csharpier` available


stevearc:
Thanks for the PR!

PR #779, opened 2025-09-24
fix(d2): format with temp file instead of stdin

KevinNitroG:
Currently d2 formatter use stdin format. It works fine with a d2 file. But with the injected formatter, it will delete the d2 content in the code block if the code block has no diff after being formatted. This PR make d2 format by filename instead of stdin.

D2 cli manual: https://d2lang.com/tour/man/

https://github.com/user-attachments/assets/4d800025-67b2-41a9-9bd5-66d3660b59ab


stevearc:
Thanks for the PR!

PR #782, opened 2025-09-28
feat: added stop_after_success option to formatter configuration

### LINKED ISSUE #717: feature request: add a flag similar to `stop_after_first` but instead it tries formatters in sequence until one succeeds

- [x] I have searched the existing issues

### Describe the feature

My understanding is `stop_after_first = true` means the `conform.nvim` will sequentially search through the list of formatters, executing the first one it finds and not searching any further.

I would propose a flag like `stop_after_success` which would make `conform.nvim` try each formatter one by one until one returned success then stop.

For simplicity sake I would assume `stop_after_first = true` is assumed when `stop_after_success = true` but that's just my preference.

### Provide background

I primarily code in rust, making heavy use of `rustfmt`. Personally I like a lot of the nightly `rustfmt` features so I often run something akin to `rustfmt +nightly`.

However due to the nature of using a nightly release sometimes it fails to handle things that the stable version could, in my experience running stable `rustfmt` if nightly fails formats the document well enough that nightly can succeed afterwards.

Ideally I would want to be able to configure `conform.nvim` to run `rustfmt +nightly` then if that fails run `rustfmt` and only report an error if both fail.

strongly desired

### Additional details

I imagined this as a flag similar to `stop_after_first` since I assume the machinery to handle such flags already exists. 

Alternatively allowing some kind of fallback field in the formatter config would also work, something like:
```lua
---@type conform.setupOpts
local opts = {
  ...
  formatters = {
    rustfmt_nightly = {
      ...
      ---@type nil
      ---| string # name of another formatter to try when this formatter fails
      ---| conform.FormatterConfigOverride # fallback config to try for the current formatter
      ---| (string|conform.FormatterConfigOverride)[] # multiple fallbacks to try in order
      fallback = "rustfmt",
      ...
    },
  },
  ...
}
```

c4patino:
- Added stop_after_success flag to formatter configuration.

The intention behind this flag is explained in the documentation. Essentially, we should attempt to run each of the formatters defined in the formatter list, and then break on the first one that successfully runs.

Resolves #717


c4patino:
> Currently we are able to expect that any time a formatter produces an error, it is an actual problem with the formatter. Introducing this functionality would change that assumption, and could impact the rest of our error handling. I would have to do a lot of reading and thinking to convince myself that our error handling, aggregation, and reporting all work correctly in all situations after this PR.

Thanks for the feedback! Yeah my usecase is definitely a bit niche, though I think that there might be other cases I'm not thinking of where it would come in useful. Essentially, we are trying to add support for `nix fmt`. Because it is built directly into the nix binaries, the formatter will _always_ be found, however if in a project without a `flake.nix`, the formatter will fail. This violates the `stop_after_first` behavior because the formatter is found but fails. The intention behind the `stop_after_success` was to allow it to continue on past the failure of `nix fmt`. 

I think there is a hybrid of the two snippets you added which might serve the use-case? Essentially a default attempt at `nix fmt`, and then on failure fall back to the defined formatters.

octvs:
I was looking around the repository for someone else with a similar issue/idea, and found the closes match on the comment above by @c4patino.

I use conform.nvim w/ several formatters on my nvim setup. But I have to disable formatting on projects that I contribute where a formatter is already defined, nearly always via `formatter` nix flake attribute. Thus I wanted conform to use `nix fmt` if a formatter is defined by the flake, if not default back to the user defined formatters.

If you could stomach a script initially generated by AI:

```lua
-------------------------------------------------------------------------------
-- conform/nix_fmt
--
-- Provides a helper for conform.nvim to detect whether a formatter is defined
-- by flake, if so delegates formatting to `nix fmt`.
--
-- Usage in your conform.nvim setup:
--
--   local nix_fmt = require("conform-nix-fmt")
--
--   require("conform").setup({
--       formatters = {
--           ...
--           nix_fmt = nix_fmt.formatter,
--       },
--       format_on_save = function(bufnr)
--           ...
--           if nix_fmt.has_nix_formatter(bufnr) then
--               return { timeout_ms = 500, formatters = { "nix_fmt" } }
--           end
--
--           return { timeout_ms = 500, lsp_format = "fallback" }
--       end,
--   })
--
-- Notes:
--   - The result of `nix eval` is cached per flake root directory.
--   - Use `:set verbose=1` to debug.
--   - To inspect the cache at runtime:
--       :lua print(vim.inspect(require("conform-nix-fmt").cache))
--   - Generated with the assistance of AI.
-------------------------------------------------------------------------------

local M = {}

local nix_formatter_cache = {}

local function dbg(msg)
    if vim.o.verbose > 0 then
        vim.notify("[conform/nix_fmt] " .. msg, vim.log.levels.DEBUG)
    end
end

local function get_nix_system()
    local uname = vim.uv.os_uname()
    local os_map = {
        Darwin = uname.machine == "arm64" and "aarch64-darwin"
            or "x86_64-darwin",
        Linux = "x86_64-linux",
    }
    local system = os_map[uname.sysname] or "x86_64-linux"
    dbg("Detected nix system: " .. system)
    return system
end

--- Check if a nix formatter is defined in the nearest flake.nix
--- Caches results per flake root directory
function M.has_nix_formatter(bufnr)
    local flake_root = vim.fs.root(bufnr, "flake.nix")
    if not flake_root then
        dbg("No flake root found, skipping nix formatter check.")
        return false
    end

    dbg("Using flake root: " .. flake_root)

    -- Return cached result if available
    if nix_formatter_cache[flake_root] ~= nil then
        dbg(
            "Cache hit for: "
                .. flake_root
                .. " -> "
                .. tostring(nix_formatter_cache[flake_root])
        )
        return nix_formatter_cache[flake_root]
    end

    dbg("Cache miss for: " .. flake_root .. ", running nix eval...")

    local system = get_nix_system()
    local cmd = string.format(
        "cd %s && nix eval .#formatter.%s --quiet --no-warn-dirty > /dev/null 2>&1; echo $?",
        vim.fn.shellescape(flake_root),
        system
    )

    local handle = io.popen(cmd)
    if not handle then
        dbg("ERROR: Failed to open pipe for command")
        nix_formatter_cache[flake_root] = false
        return false
    end
    local result = handle:read("*a")
    handle:close()

    local exit_code = tonumber(result:match("%d+"))
    dbg("Parsed exit code: " .. tostring(exit_code))

    local has_formatter = exit_code == 0
    dbg(
        "Has nix formatter: "
            .. tostring(has_formatter)
            .. " -> "
            .. (
                has_formatter and "nix fmt will run"
                or "conform will use its own formatters"
            )
    )

    nix_formatter_cache[flake_root] = has_formatter
    return has_formatter
end

--- Expose conform.nvim compatible formatter definition
M.formatter = {
    command = "nix",
    args = { "fmt", "--no-warn-dirty", "--", "$FILENAME" },
    stdin = false,
    require_cwd = true,
    cwd = require("conform.util").root_file("flake.nix"),
}

-- Expose cache for runtime inspection
M.cache = nix_formatter_cache

return M
```

The first save on a project with flake formatter is slow (the call to `nix eval`), so you might want to increase the timeout, but it works.

I wanted to share it back to here in case anyone else stumbles upon this PR with a similar search pattern to mine. Needless to say any comments & improvements would be very appreciated.

PR #786, opened 2025-10-04
feat(taplo): respect lsp options

TomJo2000:
This PR makes the `taplo` formatter aware of the config options that can be set for it[^1] via the `taplo.settings.formatter` table.

[^1]: https://taplo.tamasfe.dev/cli/usage/formatting.html

- If there is no `vim.lsp.config.taplo` table present,
(For example if `neovim/nvim-lspconfig` is not installed)
or it doesn't contain a `settings.formatter` table the formatter will gracefully fall back to the current behavior.
- If the table does exist it will be converted into `-o key=value` arguments for `taplo`'s command line arguments.

### DIFF
--- a/lua/conform/formatters/taplo.lua
+++ b/lua/conform/formatters/taplo.lua
@@ -5,5 +5,18 @@ return {
     description = "A TOML toolkit written in Rust.",
   },
   command = "taplo",
-  args = { "format", "--stdin-filepath", "$FILENAME", "-" },
+  args = function()
+    ---@type lsp.LSPObject?
+    local taplo_config = vim.tbl_get(vim.lsp.config, "taplo", "settings", "formatter")
+
+    local ret = { "format" }
+    if taplo_config ~= nil then
+      for k, v in pairs(taplo_config) do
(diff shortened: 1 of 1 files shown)


TomJo2000:
- I did some more reading and I am not 100% sure how that would interact with the file pattern rules mentioned in #771
@mmottl if you have the time I'd like your input on this.
It seems clear to me that the config file should take priority over the options parsed out of the LSP configuration in this instance.
But I suspect that since they are being supplied as command line arguments here that wouldn't be the case.
Can you come up with a reasonable way to check if an option is already being set by a config file for `taplo`?

stevearc (changes requested):
I'm a bit confused by this. If you've configured the taplo LSP, wouldn't you be using that for formatting instead of the CLI via conform? And if you haven't configured it, then why would we want to pull formatting options from there?

TomJo2000:
I disabled LSP formatting because I was having issues with conflicting settings.
I'm sure it's a bug buried somewhere in my lsp configs, but leaving the formatting to conform makes it behave properly.

TomJo2000:
Yep understandable, thanks for taking the time to review it.

PR #788, opened 2025-10-18
feat: allow tmpfile to be created in a configurable directory

### LINKED ISSUE #694: feature request: create tmpfile somewhere else

- [x] I have searched the existing issues

### Describe the feature

Create tmpfile (when using `stdin = false`) in arbitrary location

### Provide background

The problem is when Conform creates a file to lint, bundlers like Vite detect it and try to parse it, which breaks functionality in Nuxt, for example, as it sees a new "page" and tries to alter the router.

It would be nice if I could avoid telling the bundler to ignore Conform's files.

nice to have

### Additional details

_No response_

fusion44:
This enables users to redirect temporary files to an external directory, such as /tmp, thus preventing conflicts with file watchers.

I have successfully tested the changes on my system. 

Please note that lots of the tests fail on my NixOS system, but I suspect this is due to the way NixOS works, so I haven't modified any of them. The two new tests pass, though.

fixes #694

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,7 @@ require("conform").setup({
       -- When cwd is not found, don't run the formatter (default false)
       require_cwd = true,
       -- When stdin=false, use this template to generate the temporary file that gets formatted
+      -- This can be an absolute path to control where the file is created.
       tmpfile_format = ".conform.$RANDOM.$FILENAME",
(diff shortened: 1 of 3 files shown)


stevearc:
Thanks for the PR!

PR #789, opened 2025-10-20
fix(jq): use shiftwidth for indent

hmnd:
The `jq` formatter doesn't currently doesn't take shiftwidth into account when formatting. This fixes that.


stevearc:
LGTM thanks!

Issue #793, opened 2025-10-23
bug: treefmt seems to remove whole file in nix files


DockterTeagle:
### Neovim version (nvim -v)

NVIM v0.12.0-nightly+cf48741 Build type: Release LuaJIT 2.1.1741730670 Run "nvim -V1 -v" for more info

### Operating system/version

nixos

### Read debugging tips

- [x] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

- [x] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

```treefmt root directory not found.```

However, it seems that nixd formatter runs for God knows what reason, even though it is the exact same command.

### Describe the bug

I made this beast, I am more asking if anybody else is able to reproduce it. whenever using treefmt on nix files+ nixd as lsp server. treefmt or something freaks the fuck out, and deletes the whole file.

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

1. use treefmt as your everything formatter
2. have this after/lsp/nixd.lua:
```lua
---@type vim.lsp.ClientConfig
return {
  root_markers = { "flake.nix", ".git" },
  cmd = {
    "nixd",
  },
  filetypes = { "nix" },
  settings = {
    nixd = {
      inlay_hints = true,
      formatting = {
        command = { "treefmt" },
      },
      nixpkgs = {
        expr = "(builtins.getFlake (builtins.toString ./.)).inputs.nixpkgs {}",
      },
      options = {
        nixOptions = {
          expr = '(builtins.getFlake "/home/cdockter/Documents/myNixOS").nixosConfigurations.nixos.options',
        },
        home_manager = {
          expr = '(builtins.getFlake "/home/cdockter/Documents/myNixOS").homeConfigurations."cdockter".options',
        },
        flake_parts = {
          expr = "(builtins.getFlake (builtins.toString ./.)).debug.options",
        },
        perSystem = {
          expr = "(builtins.getFlake (builtins.toString ./.)).currentSystem.options",
        },
      },
    },
  },
}
```
as well as this conform option set:
```lua
  formatters_by_ft = {
    markdown = { "injected" },
    ["*"] = { "treefmt" },
  },
```
3. write seemingly to any nix file
4.  file is now blank

### Expected Behavior

I would like treefmt(the module I introduced, so I will have to fix this sometime if it is not a *massive* skill issue) to not delete my entire nix file.
[...]

PR #794, opened 2025-10-23
fix: treefmt

DrTeagle:
As mentioned in my issue, while the check for treefmt.toml makes sense, it fails when the toml file is not there, leading to undesired behavior(as is the case in nix, using either treefmt-nix or nixpkgs.treefmt.withConfig). Marking as draft so that me and @zivarah can discuss over what we should do

### DIFF
--- a/lua/conform/formatters/treefmt.lua
+++ b/lua/conform/formatters/treefmt.lua
@@ -6,6 +6,4 @@ return {
   },
   command = "treefmt",
   args = { "--stdin", "$FILENAME" },
-  require_cwd = true,
-  cwd = require("conform.util").root_file({ "treefmt.toml", ".treefmt.toml" }),
 }


stevearc:
My understanding is that `treefmt-nix` is a separate package that installs a `treefmt` command that does normal treefmt things but integrated with nix. For `treefmt-nix`, it doesn't require a toml config file because it gets its config from nix. Is that about the shape of the problem?

If that's correct, then I'd either recommend overriding the `treefmt` formatter in your user config with `require_cwd = false`, or I could potentially see the case for making another formatter configuration for `treefmt_nix`.

zivarah:
> My understanding is that `treefmt-nix` is a separate package that installs a `treefmt` command that does normal treefmt things but integrated with nix. For `treefmt-nix`, it doesn't require a toml config file because it gets its config from nix. Is that about the shape of the problem?

Yes -- specifically, `treefmt-nix` makes an executable called `treefmt` that is actually just a shell script that calls out into the real `treefmt` with a generated config file specified as an argument.

> If that's correct, then I'd either recommend overriding the `treefmt` formatter in your user config with `require_cwd = false`, or I could potentially see the case for making another formatter configuration for `treefmt_nix`.

Note that any such `treefmt_nix` formatter would need to use `flake.nix` as the root marker but also would need *something* else to understand if `treefmt-nix` is actually in use. In practice that may involve some flake evaluation depending on the implementation.

I could see a case for some more generic handling in the base `treefmt` formatter definition based on whether a `treefmt` executable exists and can be invoked without erroring. My response to #793 does give an example of how I've worked around this in my own config, though as noted there I don't know that it's upstream-ready. I'm curious for your thoughts on that.

stevearc:
That workaround seems a bit inelegant and I generally don't want to proliferate that kind of logic within conform. If there's really no easy way to tell if treefmt-nix should be used in a directory or not I might suggest that this would best be solved with `:help 'exrc'`. You can set some directory-specific logic to wire up the formatter the way you need.

PR #804, opened 2025-11-09
feat: support nested directory for tmpfile_format

PeterCardenas:
tmpfile_format with a nested directory works for the most part with the exception that the directory is left behind, leaving the file system cluttered with temporary directories. to resolve this, we remove the first directory within tmpfile_format that does not already exist.

### DIFF
--- a/lua/conform/runner.lua
+++ b/lua/conform/runner.lua
@@ -1,3 +1,4 @@
+local dir_manager = require("conform.dir_manager")
 local errors = require("conform.errors")
 local fs = require("conform.fs")
 local ft_to_ext = require("conform.ft_to_ext")
@@ -293,10 +294,6 @@ end
 ---@type table<string, boolean>
 local last_run_errored = {}
 
----Set of temporary directories to remove after formatting
----@type string[]
-local temp_dirs = {}
-
 ---@param bufnr integer
 ---@param formatter conform.FormatterInfo
 ---@param config conform.FormatterConfig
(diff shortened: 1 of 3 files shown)


lua/conform/runner.lua
```diff
@@ -375,8 +375,14 @@ local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, c
...
     uv.fs_close(fd)
     callback = util.wrap_callback(callback, function()
+      local rel_dirname = vim.fs.dirname(vim.fs.relpath(ctx.dirname, ctx.filename))
       log.debug("Cleaning up temp file %s", ctx.filename)
       uv.fs_unlink(ctx.filename)
+      if rel_dirname ~= "." then
+        local dirname = vim.fs.joinpath(ctx.dirname, rel_dirname)
+        log.debug("Cleaning up temp dir %s", dirname)
+        uv.fs_rmdir(dirname)
+      end
```
stevearc: If we're going to go to the trouble to removing dangling directories, I'd like to make sure it's a thorough solution rather than a special case for if it's a single directory deep. Since we're creating the directories just above this, it would be pretty easy to do a quick check there to keep track of which ones in that path already exist, and then we can attempt to remove the set of new directories on the way out.

lua/conform/runner.lua
```diff
@@ -370,13 +370,48 @@ local function run_formatter(bufnr, formatter, config, ctx, input_lines, opts, c
...
+    -- Thus, we check which directories relative to ctx.dirname exist and remove
+    -- the first one that does not exist.
+    local res = vim
+      .iter(
+        vim.split(
+          assert(
+            vim.fs.relpath(ctx.dirname, ctx.filename),
+            "filename " .. ctx.filename .. " is not a child of dirname " .. ctx.dirname
+          ),
+          "/",
```
stevearc: Splitting by `/` might not work on windows. Could we do this maybe procedurally or recursively using `vim.fs.dirname` to find the first parent dir that does exist?

stevearc (changes requested):

PR #807, opened 2025-11-17
fix(jq): limit indetion to [-1,7]

jspricke:
According to the documentation jq only accepts indention between -1 and
7.

Regression of 235dd79.


stevearc:
Thanks for the PR!

PR #812, opened 2025-11-23
feat(health): set conform-info filetype in health windows

mikejmcguirk:
Problem: No filetype available for the user to set options in health windows

Solution: Set "conform-info" filetype in the show_window function


stevearc:
Sure, seems fine to me

PR #818, opened 2025-12-04
feat: add Oxfmt

FLchs:
This PR add basic configuration for Oxfmt, sadly it has to use `stdin = false` until https://github.com/oxc-project/oxc/issues/14720 is resolved.


stevearc:
Thanks for the PR!

Issue #824, opened 2025-12-10
bug: Default `timeout_ms` does not work with `formatexpr()`


arkq:
### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

Debian GNU/Linux 13 (trixie)

### Read debugging tips

- [x] I have read through the [debugging tips](https://github.com/stevearc/conform.nvim/blob/master/doc/debugging.md).

### Add the debug logs

- [x] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

N/A

### Describe the bug

The `M.formatexpr = function(opts)` defined in `lua/conform/init.lua` uses custom default value for `timeout_ms`. So, when running `M.format(opts)` the default timeout from `default_format_opts` is not used.

The workaround that I have:
```lua
vim.o.formatexpr = "v:lua.myformatexpr()"
_G.myformatexpr = function()
  require("conform").formatexpr({ timeout_ms = 2000 })
end
```

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

```lua
vim.o.formatexpr = "v:lua.require('conform').formatexpr()"

require("conform").setup({
  default_format_opts = { timeout_ms = 2000 },
})
```

`gq` still timeouts after 500 ms.

### Expected Behavior

If some values are set in the `default_format_opts`, they should be used as defaults defaults in the `M.formatexpr`.

### Minimal init.lua

```Lua

```

Issue #752, opened 2025-07-16
bug: Assigning formatters to `*` breaks formatexpr fallback


fredrikfoss:
Setting a formatter for `*` breaks the formatexpr fallback. For example, if I
set formatexpr to `v:lua.require'conform'.formatexpr()`, have no formatter set
for LANG, and `*` set to use `injected`, only the `injected` formatter will run
when formatting with `gq`. It will not fall back to use the LSP formatter or
integrated formatter, as it otherwise would had I not allocated a formatter to
`*`.

Example:

```
vim.o.formatexpr = "v:lua.require'conform'.formatexpr()"

require("conform").setup({
  formatters_by_ft = {
    ["*"] = { "injected" },
  },
})

vim.lsp.enable("clangd")
```

In the above example, clangd will no longer format with `gq`. Only the injected
formatter will run.

Somewhat related, when setting `["_"] = { "FORMATTER", lsp_format = "prefer" }`,
the `lsp_format` option seems to have no effect when using formatexpr. Only
FORMATTER is run, and the running language server is ignored.

If I remember correctly, `*` and `_` are considered somewhat "controvesial" by
the plugin author, so I'm not sure these behaviors are actually wrong.

Issue #408, opened 2024-05-17
bug: timeout using gq with conform.formatexpr


chuan137:
### Neovim version (nvim -v)

NVIM v0.9.5 Build type: Release LuaJIT 2.1.1703358377

### Operating system/version

MacOs 11.5

### Add the debug logs

- [X] I have set `log_level = vim.log.levels.DEBUG` and pasted the log contents below.

### Log file

          14:41:17[DEBUG] Running formatters on /Users/d067954/SAPDevelop/ccloud/git/openstack/manila/manila/tests/share/test_api.py: { "yapf" }
          14:41:17[INFO] Run yapf on /Users/d067954/SAPDevelop/ccloud/git/openstack/manila/manila/tests/share/test_api.py
          14:41:17[DEBUG] Run command: { "yapf", "--quiet", "--lines", "137-153" }
          14:41:18[WARN] Formatter 'yapf' timeout
          14:41:18[INFO] yapf exited with code 143
          14:41:18[DEBUG] yapf stdout: { "" }
          14:41:18[DEBUG] yapf stderr: { "" }

### Describe the bug

format with 'gq' timed out, see log

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

1. set `vim.o.formatexpr = "v:lua.require'conform'.formatexpr()"` 
2. use 'gq' to format in visual mode. 

### Expected Behavior

Is it possible to set the timeout_ms option in the formatexpr function?

### Minimal init.lua

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/conform.nvim",
    config = function()
      require("conform").setup({
        log_level = vim.log.levels.DEBUG,
        -- add your config here
      })
    end,
  },
[...]

PR #837, opened 2025-12-24
Make formatexpr use the same option as conform.format()

xulongwu4:
The options for `formatexpr` is different than those used in `conform.format()`. This discrepancy causes issues when one sets up filetype-specific values for `lsp_format` or `timeout_ms`. For example, I use the following values for `formatters_by_ft`:

```lua
formatters_by_ft = {["_"] = { "trim_newlines", "trim_whitespace", lsp_format = "prefer" }}
```

I also format buffers on save.

Now when I edit a `.C` file and run `wq`, `clangd` kicks in as I set `lsp_format = "prefer"`. The file is formatted by `clangd` and saved correctly. Everything is good.

Now if I try to use `gq`, the format is performed by `trim_newlines` and `trim_whitespace`, because the default values for `lsp_format` in `formatexpr` is `fallback`. This inconsistency causes confusion when using `conform`. See also #824 #408 #752

This PR resolves this issue so that `formatexpr` will respect options used by `formatters_by_ft`.

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -832,7 +832,7 @@ M.get_formatter_info = function(formatter, bufnr)
 end
 
 M.formatexpr = function(opts)
-  -- Change the defaults slightly from conform.format
+  -- Use the same defaults as conform.format(), but force async = false and handle the range
   opts = vim.tbl_deep_extend("keep", opts or {}, {
     bufnr = vim.api.nvim_get_current_buf(),
   })


stevearc:
Hmmm...yes I think that behavior predates the `default_format_opts` and some of the other methods of configuring formatting options by filetype. Doesn't really make sense to override those values anymore. Thanks for the PR!

PR #844, opened 2026-01-02
Linking new fork of archived project, to make it easier for people

pmmalinov01:


stevearc (changes requested):
Updating the link SGTM, but this line in the README is autogenerated. You need to modify the source, which is the `meta` table in the golines formatter.

PR #854, opened 2026-01-23
add support for topiary

georgeharker:
Topiary is a tree-sitter based formatter.

I recently opened a PR there to add zsh support.

### DIFF
--- a/lua/conform/formatters/topiary.lua
+++ b/lua/conform/formatters/topiary.lua
@@ -1,0 +1,50 @@
+local util = require("conform.util")
+
+-- Map Neovim filetypes to Topiary language names
+local ft_to_language = {
+    bash = "bash",
+    sh = "bash",
+    css = "css",
+    json = "json",
+    nickel = "nickel",
+    ocaml = "ocaml",
+    rust = "rust",
+    toml = "toml",
+    query = "tree_sitter_query",
+    wit = "wit",
+    zsh = "zsh",
+}
+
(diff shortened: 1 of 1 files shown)


lua/conform/formatters/topiary.lua
```diff
@@ -0,0 +1,57 @@
...
+      "format",
+      "--language",
+      language,
+    }
+  end,
+  options = {
+    -- Override the detected language for a specific filetype
+    -- Example: { language = "bash" }
+    language = nil,
+  },
```
stevearc: Instead of having a single override value, I think it would be better to expose the `ft_to_language` map on the options like we do for the injected formatter
https://github.com/stevearc/conform.nvim/blob/5420c4b5ea0aeb99c09cfbd4fd0b70d257b44f25/lua/conform/formatters/injected.lua#L185-L187

That way it's easy for a user to override a particular language opt without clobbering all the others.

We can also fall back to using the raw vim filetype if there is no entry, and that will allow us to remove all the entries where the values are the same. I know that means the `condition` won't work anymore, but I don't think the condition is that important.

stevearc (changes requested):

lua/conform/formatters/topiary.lua
georgeharker: I'm a bit stalled with the topiary folks right now on zsh integration so this is a bit reduced in priority. Can fix as suggested or reopen when / if Zsh integration is done.

PR #862, opened 2026-03-21
fix(tex-fmt): not using tabs when prefered

snoweuph:
this annoyed me, so here Im setting the flag.


snoweuph:
@stevearc could I get a review?
or is this project dead?

stevearc:
Not dead, just have two kids and moving across the country. Thanks for the fix!

PR #863, opened 2026-03-24
feat: add pg language server config

cr4ftx:
add support for https://github.com/supabase-community/postgres-language-server

### DIFF
--- a/lua/conform/formatters/postgres-language-server.lua
+++ b/lua/conform/formatters/postgres-language-server.lua
@@ -1,0 +1,12 @@
+local util = require("conform.util")
+---@type conform.FileFormatterConfig
+return {
+  meta = {
+    url = "https://pg-language-server.com/latest/features/formatting",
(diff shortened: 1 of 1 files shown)


cr4ftx:
Hi @stevearc wdyt about adding this config?

stevearc:
conform is meant to be used with formatters that only operate as CLI tools, to provide a graceful way to integrate them with Neovim. The formatter you linked *does* expose a CLI format command, but it's a full language server. Wouldn't it make more sense for the user to just use it via LSP?

cr4ftx:
> conform is meant to be used with formatters that only operate as CLI tools, to provide a graceful way to integrate them with Neovim. The formatter you linked _does_ expose a CLI format command, but it's a full language server. Wouldn't it make more sense for the user to just use it via LSP?

I don't understand what you mean as most of the modern CLI tools can expose a LS too (biome, deno, oxc).

I could let the LS format the file but I would need to tweak my nvim config for auto format right?

Edit: adding `lsp_fallback = true` works, I don't remember why I deactivated it back then

PR #866, opened 2026-04-04
cleanup: remove old autocmd for bug fixed in 0.10

llakala:
This has been fixed for ages - see https://github.com/neovim/neovim/issues/21856#issuecomment-2104866034. Noticed this triggering in my `:set verbose` logs.

### DIFF
--- a/lua/conform/init.lua
+++ b/lua/conform/init.lua
@@ -119,16 +119,6 @@ M.setup = function(opts)
         end
       end,
     })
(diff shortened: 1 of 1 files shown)


stevearc:
LGTM, thanks!

liskin:
Should this: https://github.com/stevearc/conform.nvim/blob/dca1a190aa85f9065979ef35802fb77131911106/lua/conform/init.lua#L206-L208 be removed as well then?

PR #868, opened 2026-04-06
feat(oxfmt): Include Vite config files in formatter config

d0rianb:
Add support for Vite configuration files in formatter, to support vite plus for instance (https://viteplus.dev/guide/fmt#configuration)


stevearc:
Thanks for the PR!

PR #871, opened 2026-04-30
fix(oxfmt): align root markers with official docs

svirschevskiy:
Update oxfmt root markers according to the [docs](https://oxc.rs/docs/guide/usage/formatter/config) to include `oxfmt.config.ts`.


stevearc:
LGTM thanks!

PR #872, opened 2026-05-02
fix(prettierd): add TypeScript config files

erlliam:
Add missing Prettier TypeScript configuration files to the cwd check.


stevearc:
LGTM thanks!

PR #873, opened 2026-05-04
fix: async LSP bails on no result

liskin:
There was an inconsistency between the sync and async version of lsp_format:

* Sync version invokes all LSP clients even if they don't return a result, as long as they don't return an error.
* Async version stops as soon as no result is returned.

An example of this happening is when both pylsp and ruff are enabled, and pylsp only being used for refactoring and mypy, with the isort and black plugins disabled. In that case, pylsp gets tried first, returns no result, and conform doesn't even try ruff. It's fine with `format_on_save` (provided the formatting is fast enough), but `format_after_save` is unusable.

This commit fixes it by making the async/sync behaviour consistent.

### DIFF
--- a/lua/conform/lsp_format.lua
+++ b/lua/conform/lsp_format.lua
@@ -125,8 +125,8 @@ function M.format(options, callback)
         end
       request(client, method, params, function(err, result, ctx, _)
         vim.api.nvim_del_autocmd(auto_id)
-        if not result then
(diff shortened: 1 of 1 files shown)


liskin:
I believe the test failures weren't my fault, but I rebased nevertheless. Hopefully it'll be green now :-)

PR #880, opened 2026-05-23
fix(oxlint): accept exit code 1 as success

rickharris:
Prior to this commit, the oxlint formatter failed to format files with both fixable and unfixable errors. The unfixable errors would cause oxlint to exit with exit code 1 and conform would report a failure to format. With this change files with both fixable and unfixable errors are able to be formatted successfully.


stevearc:
LGTM thanks!

PR #881, opened 2026-05-26
actually ignore interpolations

auscyber:
This ignores interpolations inside of injected text and replaces them with placeholders so they can be formatted correctly. one example of this being useful is when using say nix with injected lua or bash and stylua will error out when it sees an interpolation like ${var} inside a string. This PR turns it something like __CONFORM_PLACEHOLDER_idx_ to make sure that it can be formatted properly. the actual replacement is not done with simply replace and is actually using the treesitter tree as they include information of injections. i've also made it possible to add your own queries for replacement to make this work as such

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ tests/fake_formatter_output
 scripts/nvim_doc_tools
 scripts/nvim-typecheck-action
 tmp/
+.workspaces/
--- a/doc/formatter_options.md
+++ b/doc/formatter_options.md
@@ -54,6 +54,14 @@ options = {
     teal = "tl",
     typescript = "ts",
   },
+  interpolation_queries = {
+    nix = "(interpolation) @interp",
+    -- JS/TS template strings contain `${ ... }` substitutions.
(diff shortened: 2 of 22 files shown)


stevearc:
Please mark this as draft until it's ready to review. If it is ready to review, please include a description in the PR.

auscyber:
> Please mark this as draft until it's ready to review. If it is ready to review, please include a description in the PR.



> This ignores interpolations inside of injected text and replaces them with placeholders so they can be formatted correctly. one example of this being useful is when using say nix with injected lua or bash and stylua will error out when it sees an interpolation like ${var} inside a string. This PR turns it something like __CONFORM_PLACEHOLDER_idx_ to make sure that it can be formatted properly. the actual replacement is not done with simply replace and is actually using the treesitter tree as they include information of injections. i've also made it possible to add your own queries for replacement to make this work as such

done

c-y-p-e:
This is fantastic, finally made it possible to enable injected formatting for my nix files, thanks!

PR #883, opened 2026-06-12
fix(formatters/ansible-lint): add missing exit code

lkhn:
When the --fix option is used, ansible-lint returns 8 as a exit code after performing auto-fixes [1].

Include 8 into the expected exit codes that indicate success.

[1]: https://github.com/ansible/ansible-lint/blob/4b79c996172bb941953e5c4287535378e232c16f/src/ansiblelint/constants.py#L30-L41


stevearc:
LGTM thanks for the PR!

PR #895, opened 2026-08-10
fix(texfmt): set tabsize to 1 if expandtab is true

antinomie8:
When indenting with tabs, passing `shiftwidth` as `--tabsize` will result in this many tabs being used for indentation, instead of a single tab. So if a tab is configured to be rendered as 4 spaces, then each line will be 4 times as much indented as it should.


stevearc:
LGTM thanks!

